PR #3, opened 2017-06-29
Test fail improvements [2890239]

christopher-hopper:
## Suggested Changes

Ensure our CI server, and other developers collaborating on the project, runs and tests on the same dependencies.

**Note:** This change will not have any effect on other projects that depend on this library.

For details on the effect of comitting the `composer.lock` file to version control, see the following documentation:

- [Composer - Libraries - Lock file](https://getcomposer.org/doc/02-libraries.md#lock-file)
- [Composer - Basic usage - Commit Your `composer.lock` File to Version Control](https://getcomposer.org/doc/01-basic-usage.md#commit-your-composer-lock-file-to-version-control)

## Related Issues

- [#2890239 PHPUnit Tests are failing](https://www.drupal.org/node/2890239)
- [#2885266 Drop support for PHP 5.4 and 5.5](https://www.drupal.org/node/2885266)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -6,3 +6,4 @@
 /phpcs.xml.dist
 /phpunit.xml.dist
 /vendor export-ignore
+/composer.lock export-ignore
--- a/.gitattributes
+++ b/.gitattributes
@@ -4,6 +4,6 @@
 /README.md export-ignore
 /coder_sniffer/Drupal/Test export-ignore
 /coder_sniffer/DrupalPractice/Test export-ignore
-/phpcs.xml.dist
-/phpunit.xml.dist
+/phpcs.xml.dist export-ignore
+/phpunit.xml.dist export-ignore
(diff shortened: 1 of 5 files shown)


pfrenssen:
No we shouldn't commit the lock file otherwise we will not be alerted if something breaks. If we commit the lock file then we are oblivious to incompatibilities that arise with our own dependencies, but our users aren't, since the versions in the lock file do not propagate to our user's projects.

If any incompatibility will arise then we will not know, but our users will still be affected. That is a bad idea.

christopher-hopper:
@pfrenssen after a few beers and a discussion at the pub I tend to agree with you.

I agree it's wiser to let the tests surface incompatibilities as underlying libraries change, due to the way this all works. Perhaps Travis.ci should run the tests weekly, if no commits are merged to the base branch, and a status badge could be added somewhere.

PR #4, opened 2017-08-28
Issue #2902536: Add $config in $coreGlobals from Drupal.NamingConventions.ValidGlobal.

mfernea:


pfrenssen (approved):
As pointed out by @mfernea in the drupal.org issue, `global $config` is listed in `globals.api.php` so this is a valid addition to the whitelist.

PR #6, opened 2017-08-29
Issue #2905348: Add .gitignore file.

mfernea:


wundo:
This has been marked as won't fix in DO, can we close it here as well?

mfernea:
You're right. I'm closing the PR.

PR #8, opened 2017-09-06
Issue #2906931:  Yield should be a recognised return statement.

mfernea:


mfernea:
The test is failing on PHP 5.4 because yield was introduced in PHP 5.5 and PHPCS_T_YIELD is not available hence T_YIELD is not available too. I don't know how to fix this.

PR #9, opened 2018-04-11
Update to phpcs 3.2.3

fenetikm:
Fixed up a couple of sniffs that had issues, removed the obsolete DrupalPractice project tests. All tests passed locally.


fenetikm:
I can see that the phpunit tests pass but the phpcs testing doesn't - did it ever work? The code in the module looks a bit behind current D8 coding standards. Could we get this merged and then create another issue around coding standards?

fenetikm:
Actually, I can see the problem is the rule in phpcs xml dist, will sort it out.

PR #16, opened 2018-09-27
Remove older and unsupported PHP versions

truls1502:
https://www.drupal.org/project/coder/issues/3002772

> I recommend removing unsupported PHP version for the Drupal part such as PHP 5.4 and 5.5.
> 
> I know it is published an explanation on https://github.com/squizlabs/PHP_CodeSniffer/releases/tag/3.0.0a1
> 
> However, please give your opinion here if we should still keep PHP 5.4 and 5.5. Since I am trying to follow https://www.drupal.org/docs/8/system-requirements/php-requirements

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,6 @@ sudo: false
 matrix:
   fast_finish: true
   include:
-    - php: 5.4
-    - php: 5.5
     - php: 5.6
     - php: 7.0
     - php: 7.1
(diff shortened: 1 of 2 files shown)


pfrenssen:
Drupal 7 still supports these PHP versions. There are still a large number of Drupal 7 sites using Coder so I would like to keep these versions tested for as long as Drupal 7 is still supported.

If there would be another pressing reason then we can reconsider this. For example if PHP_CodeSniffer would drop support for them we will follow suit. But for the moment we have very little disadvantages from testing and supporting these versions. It's just that the tests take a longer time to run, but that's a small sacrifice.

PR #22, opened 2019-03-16
tests(phpcs): Update PHPCS to 3.4.1 and fix phpunit tests

klausi:
Issue: https://www.drupal.org/project/coder/issues/3040784

PHPIUnit tests are broken, if we update PHPCS then they are fixed again.

This is wasting resources because a git clone of PHP_CodeSniffer is 200MB, so I will open an issue there that they should release 3.4.1

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
-dist: xenial
+# For PHP 5.5 and lower testing we need to stick to Ubuntu 14.04.
+dist: trusty
 language: php
 sudo: false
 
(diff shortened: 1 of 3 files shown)


klausi:
Updated the Travis Config a bit to enable composer install caching. Should reduce the random composer install fails from above.

klausi:
The composer caching settings have brought down the Travis test time now again, even if we use the PHPCS dev version for now.

truls1502 (approved):

pfrenssen (approved):

PR #23, opened 2019-03-30
tests(ProjectDetection): Fix disabled project file tests

klausi:
See https://www.drupal.org/project/coder/issues/2962880

Tests will fail because we need #22 first.

### DIFF
--- a/coder_sniffer/DrupalPractice/Test/ProjectDetection/ProjectUnitTest.php
+++ b/coder_sniffer/DrupalPractice/Test/ProjectDetection/ProjectUnitTest.php
@@ -3,7 +3,6 @@
 namespace DrupalPractice\ProjectDetection;
 
 use DrupalPractice\Project;
-use PHP_CodeSniffer\Files\File;
 
 /**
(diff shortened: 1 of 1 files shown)


jonathan1055:
Did you notice that the failure of build 64.1 (PHP5.4) was for a parse/syntax error. The failures for builds 64.2 and all the rest were the expected "number of css error count wrong". So the new syntax you added on line 27 is not compatible with PHP5.4. Is this a problem? Are we intending to support running at PHP5.4?

klausi:
Good point! We could simply drop PHP 5.4 support, but that is probably a separate issue.

I will update this PR to not use ::class syntax.

pfrenssen:
I am in favor to drop 5.4 support but this might break user's CI pipelines, so we should probably hold this off until the next major release. We can then even decide to drop all unsupported PHP versions.

But for now yes let's try to keep 5.4 support going.

klausi:
Opened https://www.drupal.org/project/coder/issues/3047617 to drop PHP 5 support.

PR #30, opened 2019-04-14
Yield should be a recognised return statement

klausi:
Testing #8 again after we dropped PHP 5.4 support.


klausi:
Works, merging the original now.

PR #34, opened 2019-04-23
Project version is not correctly matched in deprecated checks.

jsacksick:
See https://www.drupal.org/project/coder/issues/3050166


jonathan1055:
This looks good to me. The additions 'good' test case shows it is working.

PR #35, opened 2019-04-23
Add a DrupalPractice rule for $strictConfigSchema in tests

jhedstrom:
This is the corresponding PR for https://www.drupal.org/project/coder/issues/2957116 and is now targeted against the 3.x branch.

### DIFF
--- a/coder_sniffer/DrupalPractice/Sniffs/Objects/StrictSchemaDisabledSniff.php
+++ b/coder_sniffer/DrupalPractice/Sniffs/Objects/StrictSchemaDisabledSniff.php
@@ -79,14 +79,9 @@ class StrictSchemaDisabledSniff extends AbstractVariableSniff
     protected function isTestClass(File $phpcsFile, $stackPtr)
     {
         // Only applies to test classes, which have Test in the name.
-        $types  = array(
-                   T_CLASS,
-                   T_TRAIT,
-                  );
-        $class  = $phpcsFile->findPrevious($types, ($stackPtr - 1));
-        $name   = $phpcsFile->findNext([T_STRING], $class);
-        $tokens = $phpcsFile->getTokens();
-
+        $tokens   = $phpcsFile->getTokens();
+        $classPtr = key($tokens[$stackPtr]['conditions']);
+        $name     = $phpcsFile->findNext([T_STRING], $classPtr);
         return strpos($tokens[$name]['content'], 'Test') !== false;
 
     }//end isTestClass()
--- a/coder_sniffer/DrupalPractice/Sniffs/Objects/StrictSchemaDisabledSniff.php
+++ b/coder_sniffer/DrupalPractice/Sniffs/Objects/StrictSchemaDisabledSniff.php
@@ -1,0 +1,135 @@
+<?php
+/**
+ * DrupalPractice_Sniffs_Objects_StrictSchemaDisabledSniff.
+ *
+ * @category PHP
+ * @package  PHP_CodeSniffer
+ * @link     http://pear.php.net/package/PHP_CodeSniffer
+ */
+
+namespace DrupalPractice\Sniffs\Objects;
+
+use PHP_CodeSniffer\Files\File;
+use PHP_CodeSniffer\Sniffs\AbstractVariableSniff;
+
+/**
+ * Checks that $strictConfigSchema is not set to FALSE in test classes.
+ *
+ * @category PHP
+ * @package  PHP_CodeSniffer
+ * @link     http://pear.php.net/package/PHP_CodeSniffer
+ */
+class StrictSchemaDisabledSniff extends AbstractVariableSniff
+{
+
+    /**
+     * The name of the variable in the test base class to disable config schema checking.
+     */
+    const STRICT_CONFIG_SCHEMA_NAME = '$strictConfigSchema';
+
+
+    /**
+     * Processes this test, when one of its tokens is encountered.
+     *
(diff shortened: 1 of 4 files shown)


coder_sniffer/DrupalPractice/Sniffs/Objects/StrictSchemaDisabledSniff.php
```diff
@@ -0,0 +1,121 @@
...
+use PHP_CodeSniffer\Sniffs\AbstractScopeSniff;
+
+/**
+ * Checks that $strictConfigSchema is not set to FALSE in test classes.
+ *
+ * @category PHP
+ * @package  PHP_CodeSniffer
+ * @link     http://pear.php.net/package/PHP_CodeSniffer
+ */
+class StrictSchemaDisabledSniff extends AbstractScopeSniff
```
klausi: Hm, are you sure AbstractScopeSniff is the correct class and not AbstractVariableSniff with processMemberVar()?

What happens if you have a variable named $strictConfigSchema as local variable in a class method? Please add a test case for that so that we see that this sniff does not trigger there.

klausi (changes requested):

coder_sniffer/DrupalPractice/Sniffs/Objects/StrictSchemaDisabledSniff.php
jhedstrom: Good catch! That's more along the lines of the 2.x version of this I made, but I missed that base class.

jhedstrom:
I've updated the PR--not sure how to remove the 'Changes requested' bit.

klausi:
You can request a new review, see https://github.blog/changelog/2019-02-21-re-request-review-on-a-pull-request/

Hm, please don't force-push pull requests because now I don't have an interdiff of what you changed recently. I will squash-merge pull requests anyway, so in the end only 1 commit will end up on the 8.x-3.x branch in any case.

jhedstrom:
Ah, good to know regarding commit squashing--I was just trying to keep the commit history simple :)

coder_sniffer/DrupalPractice/Test/Objects/StrictSchemaDisabledUnitTest.inc
```diff
@@ -0,0 +1,40 @@
...
+     * Disable strict config checking during testing.
+     */
+    protected $strictConfigSchema = FALSE;
+}
+
+class TestNull {
+    protected $strictConfigSchema = NULL;
+
+    public function strictConfigSchema()
+    {
```
klausi: Sorry, missed this in my first review: can you use Drupal coding standards in the test files? 2 space indentation, braces on the line before etc.

If you create StrictSchemaDisabledUnitTest.inc.fixed then you see what it should look like :-)

coder_sniffer/DrupalPractice/Sniffs/Objects/StrictSchemaDisabledSniff.php
```diff
@@ -0,0 +1,135 @@
...
+     *   Returns TRUE if the current class is a test class.
+     */
+    protected function isTestClass(File $phpcsFile, $stackPtr)
+    {
+        // Only applies to test classes, which have Test in the name.
+        $types  = array(
+                   T_CLASS,
+                   T_TRAIT,
+                  );
+        $class  = $phpcsFile->findPrevious($types, ($stackPtr - 1));
```
klausi: Instead of searching we can use something like `$classPtr    = key($tokens[$stackPtr]['conditions']);` as in ValidVariableNameSniff.

coder_sniffer/DrupalPractice/Sniffs/Objects/StrictSchemaDisabledSniff.php
```diff
@@ -0,0 +1,135 @@
...
+        $tokens = $phpcsFile->getTokens();
+
+        if (($tokens[$stackPtr]['content'] === static::STRICT_CONFIG_SCHEMA_NAME) && ($this->isTestClass($phpcsFile, $stackPtr) === true)) {
+            $find = array(
+                     T_FALSE,
+                     T_TRUE,
+                     T_NULL,
+                     T_SEMICOLON,
+                    );
+            $next = $phpcsFile->findNext($find, ($stackPtr + 1));
```
klausi, line 51: This can go wrong if a class declares variables like `protected $foo, $bar = 5;`. So I think you also need to search for ','. We should also add a test case for this.

klausi (changes requested):
I guess it is not worth it to implement a fixer with this, so we are almost done here. I noticed some small stuff that should be improved.

coder_sniffer/DrupalPractice/Sniffs/Objects/StrictSchemaDisabledSniff.php
jhedstrom: I added a test for it, but didn't need to update the code since it eventually ends with a semi-colon regardless.

coder_sniffer/DrupalPractice/Sniffs/Objects/StrictSchemaDisabledSniff.php
jhedstrom: I changed this to 

```
        // Only applies to test classes, which have Test in the name.
        $tokens   = $phpcsFile->getTokens();
        $classPtr = key($tokens[$stackPtr]['conditions']);
        $name     = $phpcsFile->findNext([T_STRING], $classPtr);
        return strpos($tokens[$name]['content'], 'Test') !== false;
```

PR #43, opened 2019-05-05
chore(ScopeIndent): Sync upstream changes from PHPCS

klausi:
While fixing coding standards I noticed that this sniff is outdated from upstream, so I synced it again.

I will run this on Drupal core to see if it flags something unusual there.


klausi:
Opened https://github.com/squizlabs/PHP_CodeSniffer/pull/2502

PR #46, opened 2019-05-23
fix(Objects) Detect use of global functions in classes that have access to the container

pfrenssen:
Fixes https://www.drupal.org/project/coder/issues/3056538


pfrenssen:
Note that this does not yet cover the case of classes that have access to the container because they _extend_ a class that implements `ContainerInjectionInterface`. It only covers classes that directly implement the interface. I looked for a way in PHP_CodeSniffer to find a full list of interfaces that is implemented by the class hierarchy but it seems this does not exist yet.

klausi:
Looks good, thanks!

PR #47, opened 2019-06-06
fix for FunctionCommentSniff: Support object type hint for php 7.2 and greater

pwolanin:
See drupal.org #3059698

There is some related code already in squizlabs:

squizlabs/php_codesniffer/src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php
```
                if ($this->phpVersion >= 70200) {
                    if ($suggestedName === 'object') {
                        $suggestedTypeHint = 'object';
                    }
                }
```

### DIFF
--- a/coder_sniffer/Drupal/Test/Commenting/FunctionCommentUnitTest.php
+++ b/coder_sniffer/Drupal/Test/Commenting/FunctionCommentUnitTest.php
@@ -82,12 +82,12 @@ class FunctionCommentUnitTest extends CoderSniffUnitTest
 
         case 'FunctionCommentUnitTest.3.inc':
             if (PHP_VERSION_ID < 70200) {
-              return [
-                  9 => 1,
-                  12 => 1,
-              ];
+                return [
+                    9  => 1,
+                    12 => 1,
+                ];
             } else {
-              return [9 => 1];
+                return [9 => 1];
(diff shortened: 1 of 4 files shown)


coder_sniffer/Drupal/Test/Commenting/FunctionCommentUnitTest.php
```diff
@@ -79,6 +79,16 @@ public function getErrorList($testFile=null)
 
         case 'FunctionCommentUnitTest.2.inc':
             return [8 => 1];
+
+        case 'FunctionCommentUnitTest.3.inc':
+            if (PHP_VERSION_ID < 70200) {
+                return [
+                    9  => 1,
+                    14 => 1,
```
klausi, line 87: I think it is not a good idea to have different test results per PHP version. Line 14 should never be flagged as error if the PHP type hint is object. We should just make "object" an acceptable type hint. People might run PHPCS on a lower PHP version and then get different results than on their CI or prod enviorinment. It is just an invitation for confusion.

klausi (changes requested):
Thanks a lot! I still think we should not do the PHP version checking, can you change this to always accept "object"?

klausi:
Committed a slightly different version, thanks!

PR #56, opened 2019-08-12
Avoid running ClassFileNameSniff on text files

eiriksm:
Ref: https://www.drupal.org/project/coder/issues/3074186

### DIFF
--- a/coder_sniffer/Drupal/Sniffs/Classes/ClassFileNameSniff.php
+++ b/coder_sniffer/Drupal/Sniffs/Classes/ClassFileNameSniff.php
@@ -57,6 +57,20 @@ class ClassFileNameSniff implements Sniff
             return ($phpcsFile->numTokens + 1);
         }
 
+        // If the file is not a known php file (like a txt file or a md file),
+        // we do not need to check if the file is named after the class we
+        // found.
+        $extension     = pathinfo($fullPath, PATHINFO_EXTENSION);
+        $phpExtensions = [
+            'inc',
+            'php',
+            'module',
+            'theme',
+        ];
+        if (in_array($extension, $phpExtensions) === false) {
+            return ($phpcsFile->numTokens + 1);
+        }
+
         $tokens  = $phpcsFile->getTokens();
         $decName = $phpcsFile->findNext(Tokens::$emptyTokens, ($stackPtr + 1), null, true);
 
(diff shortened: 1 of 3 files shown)


coder_sniffer/Drupal/Sniffs/Classes/ClassFileNameSniff.php
```diff
@@ -68,6 +68,17 @@ public function process(File $phpcsFile, $stackPtr)
             $fileName = str_replace('_', '', ucwords($fileName, '_')).'SubContext';
         }
 
+        // If the file is either txt or md, it probably should not be named
+        // after the class we are searching for.
+        $extension         = pathinfo($fullPath, PATHINFO_EXTENSION);
+        $ignoredExtensions = [
```
klausi: Now that I think about it - I think we should not have a list of extensions to ignore, we should reverse this and only accepts *.php files.

Can you change that and move this check further up? Right after where we check for empty/no file name.

klausi (changes requested):

coder_sniffer/Drupal/Sniffs/Classes/ClassFileNameSniff.php
eiriksm: I was thinking we could also include other known php files (like .module, .inc and .theme) since the fact that you are declaring a class in there might be a code smell. I updated the code with that, what do you think?

eiriksm: ...and also the fact that we have code specifically handling .behat.inc files in both text fixtures and in the sniff class :)

coder_sniffer/Drupal/Sniffs/Classes/ClassFileNameSniff.php
```diff
@@ -57,6 +57,20 @@ public function process(File $phpcsFile, $stackPtr)
...
 
+        // If the file is not a known php file (like a txt file or a md file),
+        // we do not need to check if the file is named after the class we
+        // found.
+        $extension     = pathinfo($fullPath, PATHINFO_EXTENSION);
+        $phpExtensions = [
+            'inc',
+            'php',
+            'module',
+            'theme',
```
klausi: I think we should only check *.php files. Those are the only files where we want to compare file names with class names.

Then we can also remove the special handling for behat.inc files below. It was never the intention to test special behat files with this sniff anyway. This is for PSR4 *.php files and their names.

klausi (changes requested):
Thank you, but let's just check *.php files. Can you change that?

eiriksm:
Updated the branch and fixed comments :)

PR #59, opened 2019-09-09
Use correct runtimeexception class in unit test class

eiriksm:


eiriksm:
See issue https://www.drupal.org/project/coder/issues/3080158

PR #64, opened 2019-11-28
Issue #3057988 Shorter @deprecated and @trigger_error sniff names

jonathan1055:
Drupal issue https://www.drupal.org/project/coder/issues/3057988#comment-13370817

Synchronise sniff names and remove unnecessary text. No actual coding changes.


jonathan1055:
The tests fail at PHP7.1 and 7.2 due to needed PHPUNIT6 (I think). The tests at lower PHP versions are fine. The d.o. issue for this is https://www.drupal.org/project/coder/issues/3097302

klausi:
Thanks for tracking that down, then let's fix the test execution first in that issue.

klausi:
Sorry, just realized now that we can't change those names now because they are already in a stable release of Coder (8.3.6). Changing them now would mean the next release would break people's PHPCS configuration where they for example silenced those sniffs.

We could shorten them in Coder 9, but to be honest it does not seem worth the config breaking change. So I think we should stick with those slightly longer names.

jonathan1055:
OK thanks for the info. Yes it would break BC. These rules are not used by Core yet, but I had forgotten that contrib may have used them. Although a quick search on http://grep.xnddx.ru/search?text=Drupal.Commenting only shows a few contrib project usages and none are for the sniffs I am talking about.

Issue #54, opened 2019-07-16
DrupalPractice can use wrong autoloader


danepowell:
https://github.com/pfrenssen/coder/pull/29 introduced the following line into the DrupalPractice ruleset:
```
<autoload>../Drupal/autoload.php</autoload>
```

This can cause phpcs to break in rather exciting and unexpected ways depending on the architecture of the host system.

In my case, I have a package (BLT) that depends on coder and uses phpcs to do internal code sniffing. I also happen to develop on Drupal core, although this shouldn't really be relevant (but you'll see why it is momentarily).

On my Mac OSX machine, I keep all packages under development in a `~/packages` directory:
```
/Users/dane/packages/blt
/Users/dane/packages/drupal
(etc...)
```

When I was developing on BLT today, I ran `./vendor/bin/phpcs` in the `/Users/dane/packages/blt` directory and was greeted with the following error:
```
require(/Users/dane/packages/Drupal/vendor/autoload.php): failed to open stream: No such file or directory in /Users/dane/packages/Drupal/autoload.php
```

This is unexpected because the drupal package should be completely isolated from BLT, as it's in a sibling directory!

I think what's happening here is that because of that definition in the ruleset, _and_ the fact that OSX is case-insensitive, _and_ I just happen to have Drupal cloned in a sibling directory, Composer tries to autoload from that directory and explodes.

claudiu-cristea:
I'm getting something similar on MacOS but also with `DrupalPractice`:

```bash
$ ./vendor/bin/phpcs --standard=DrupalPractice
PHP Fatal error:  Cannot redeclare composerRequireDrupal8() (previously declared in /Users/clau/development/8/time/vendor/composer/autoload_real.php:67) in /Users/clau/development/8/Drupal/vendor/composer/autoload_real.php on line 74
```

EDIT: And, yes, I also have a `/Users/clau/development/8/drupal` directory.

michael-wojcik:
I ran into the same issue just now, trying to use `DrupalPractice` with `phpcs`.

I've created a PR that should fix this issue for anyone still running into this: https://github.com/pfrenssen/coder/pull/70

klausi:
Looks like a PHPCS autoloading big, opened https://github.com/squizlabs/PHP_CodeSniffer/pull/2751

PR #70, opened 2019-12-05
#54: Fixes issue with phpcs using wrong autoloader

michael-wojcik:
Patch to fix issue mentioned in https://github.com/pfrenssen/coder/issues/54


danepowell (approved):
Awesome, thanks! I can confirm that this fixes the problem and makes sense as a fix. It's still not 100% deterministic, but it's way less likely to cause conflicts than the old path.

klausi:
Thanks, used a bit of a different approach in #72

PR #71, opened 2019-12-06
Add configurable line limit to Drupal.Arrays.Array

arkener:
Reference: https://www.drupal.org/project/coder/issues/3099288


coder_sniffer/Drupal/Test/Arrays/ArrayUnitTest.php
```diff
@@ -35,23 +45,47 @@ public function getErrorList()
...
+
+
+    /**
+     * Returns a list of test files that should be checked.
+     *
+     * @param string $testFileBase The base path that the unit tests files will have.
+     *
+     * @return array The list of test files.
+     */
+    protected function getTestFiles($testFileBase)
```
klausi: no need to override this method. The parent will find *1.inc and *2.inc just fine. See for example DocCommentStarUnitTest

klausi (changes requested):
Thanks, the approach looks good. We just need to clean up the test case a bit.

No need to rename ArrayUnitTest.inc, see how we deal with DocCommentStarUnitTest with multiple files for example.

coder_sniffer/Drupal/Test/Arrays/ArrayUnitTest.php
arkener: Thanks for the feedback. I've reverted the change to the existing array unit tests and removed unneeded overrides

PR #81, opened 2020-01-17
Recognise 'yield from' as valid return statement

arkener:
Reference: https://www.drupal.org/project/coder/issues/3106904

Also checked if any other sniffs currently used the T_YIELD token, this was not the case.

### DIFF
--- a/coder_sniffer/Drupal/Test/Commenting/FunctionCommentUnitTest.inc
+++ b/coder_sniffer/Drupal/Test/Commenting/FunctionCommentUnitTest.inc
@@ -506,3 +506,15 @@ class Small {
 function test37(array $matches, array $sub_key, $to) {
 
 }
+
+/**
+ * Yield from should be a recognised return statement.
+ *
+ * @return int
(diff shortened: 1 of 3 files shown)


coder_sniffer/Drupal/Test/Commenting/FunctionCommentUnitTest.inc
```diff
@@ -506,3 +506,15 @@ class Small {
...
+
+/**
+ * Yield from should be a recognised return statement.
+ *
+ * @return int
+ *   Integer value.
+ */
+function test38($a, $b) {
+  for ($i = 1; $i <= 3; $i++) {
+    yield from $i;
```
klausi: This is confusing because $i is an integer. `yield from` should be used on arrays or iterators. Can we have an example that yields from an array instead?

arkener: Good point, I've updated the test function to a proper yield from implementation.

PR #85, opened 2020-01-29
Add check if the inheritdoc tag is the only line of a docblock

arkener:
Reference: https://www.drupal.org/project/coder/issues/3109921


klausi:
Sorry, we won't be doing this, see the comments in the issue. Thanks anyway!

PR #86, opened 2020-02-02
Fix incorrect comment for CoderSniffUnitTest::checkAllSniffCodes()

arkener:
The CoderSniffUnitTest::checkAllSniffCodes() method currently states the following:
_False if just the current sniff should be checked, false if all sniffs should be checked._
This should be:
_False if just the current sniff should be checked, true if all sniffs should be checked._


arkener:
Tests are currently broken because of an incompatibility with PHP_CodeSniffer 3.5.4, see #88

klausi:
Thank you!

PR #87, opened 2020-02-02
feat(VariableAnalysis): Add support for short-hand list unpacking

arkener:
Reference: https://www.drupal.org/project/coder/issues/2876245


arkener:
Tests are currently broken because of an incompatibility with PHP_CodeSniffer 3.5.4, see #88

PR #89, opened 2020-02-09
feat(FullyQualifiedNamespace): Add alias support and avoid classname conflicts

arkener:
Reference: https://www.drupal.org/project/coder/issues/3112316

### DIFF
--- a/coder_sniffer/Drupal/Test/Classes/FullyQualifiedNamespaceUnitTest.inc.fixed
+++ b/coder_sniffer/Drupal/Test/Classes/FullyQualifiedNamespaceUnitTest.inc.fixed
@@ -2,8 +2,66 @@
 
 /**
  * @file
+ * Example.
  */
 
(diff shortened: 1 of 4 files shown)


coder_sniffer/Drupal/Test/Classes/FullyQualifiedNamespaceUnitTest.inc.fixed
```diff
@@ -2,8 +2,66 @@
...
+   * Description.
+   */
+  public function test6(TestDiff\Bar $bar) {
+
+  }
+
+  /**
+   * Description.
+   */
+  public function test7(TestDiff\TestAlias $anotherAlias) {
```
klausi, line 63: we report an error here, but we do not fix it. I guess we could, but let's leave this out of scope for this fix for now.

klausi (approved):

PR #90, opened 2020-02-23
Issue 3057988 Automatically fix deprecated layout text

jonathan1055:
https://www.drupal.org/project/coder/issues/3057988

### DIFF
--- a/coder_sniffer/Drupal/Sniffs/Commenting/DeprecatedSniff.php
+++ b/coder_sniffer/Drupal/Sniffs/Commenting/DeprecatedSniff.php
@@ -11,9 +11,10 @@ namespace Drupal\Sniffs\Commenting;
 
 use PHP_CodeSniffer\Files\File;
 use PHP_CodeSniffer\Sniffs\Sniff;
+use PHP_CodeSniffer\Config;
 
 /**
- * Ensures standard format of a @deprecated text.
+ * Ensures standard format of @ deprecated tag text in docblock.
  *
  * @category PHP
  * @package  PHP_CodeSniffer
@@ -21,6 +22,15 @@ use PHP_CodeSniffer\Sniffs\Sniff;
  */
 class DeprecatedSniff implements Sniff
 {
+
+    /**
+     * Show debug output for this sniff.
+     *
+     * Use phpcs --runtime-set deprecated_debug true
+     *
+     * @var boolean
+     */
+    private $debug = false;
 
 
     /**
@@ -30,6 +40,10 @@ class DeprecatedSniff implements Sniff
      */
     public function register()
     {
+        if (defined('PHP_CODESNIFFER_IN_TESTS') === true) {
+            $this->debug = false;
+        }
+
         return [T_DOC_COMMENT_TAG];
 
     }//end register()
@@ -46,6 +60,11 @@ class DeprecatedSniff implements Sniff
      */
     public function process(File $phpcsFile, $stackPtr)
     {
+        $debug = Config::getConfigData('deprecated_debug');
+        if ($debug !== null) {
+            $this->debug = (bool) $debug;
+        }
+
         $tokens = $phpcsFile->getTokens();
 
         // Only process @deprecated tags.
@@ -55,6 +74,7 @@ class DeprecatedSniff implements Sniff
 
         // Get the end point of the comment block which has the deprecated tag.
         $commentEnd = $phpcsFile->findNext(T_DOC_COMMENT_CLOSE_TAG, ($stackPtr + 1));
+
         // Get the full @deprecated text which may cover multiple lines.
         $textItems = [];
         $lastLine  = $tokens[($stackPtr + 1)]['line'];
@@ -89,7 +109,7 @@ class DeprecatedSniff implements Sniff
         if (count($matches) !== 4) {
             // The full text does not match the standard. Try to find fixes by
             // testing with a relaxed set of criteria, based on common
-            // formatting variations. This is designed for Core fixes.
+            // formatting variations. This is designed for Core fixes only.
             $error = "The text '@deprecated %s' does not match the standard format: ".$standardFormat;
             // All of the standard text should be on the first comment line, so
             // try to match with common formatting errors to allow an automatic
@@ -100,19 +120,23 @@ class DeprecatedSniff implements Sniff
                 // Get just the first line of the text.
                 $key   = array_keys($textItems)[0];
                 $text1 = $textItems[$key];
+                // Matching on (drupal|) here says that we are only attempting to provide
+                // automatic fixes for Drupal core, and if the project is missing we are
+                // assuming it is Drupal core. Deprecations for contrib projects are much
+                // less frequent and faults can be corrected manually.
                 preg_match('/^(.*)(as of|in) (drupal|)( |:|)+([\d\.\-xdev\?]+)(,| |. |)(.*)(removed|removal)([ |from|before|in|the]*) (drupal|)( |:|)([\d\-\.xdev]+)( |,|$)+(?:release|)(?:[\.,])*(.*)$/i', $text1, $matchesFix);
 
                 if (count($matchesFix) >= 12) {
-                    // This problem is a drupal core file and is fixable.
-                    if (empty($matchesFix[1]) === false) {
-                        // Verify that its acceptable to remove the text in [1].
-                        echo('>>> File: '.$phpcsFile->path.' line '.$tokens[($stackPtr)]['line'].PHP_EOL);
-                        echo('>>> First line: '.$text1.PHP_EOL);
-                        echo('>>> Would remove: '.$matchesFix[1].PHP_EOL);
+                    // It is a Drupal core deprecation and is fixable.
+                    if (empty($matchesFix[1]) === false && $this->debug === true) {
+                        // For info, to check it is acceptable to remove the text in [1].
+                        echo('DEBUG: File: '.$phpcsFile->path.', line '.$tokens[($stackPtr)]['line'].PHP_EOL);
+                        echo('DEBUG: "@deprecated '.$text1.'"'.PHP_EOL);
+                        echo('DEBUG: Fix will remove: "'.$matchesFix[1].'"'.PHP_EOL);
                     }
 
-                    $ver1 = str_Replace(['x-dev', 'x'], ['0', '0'], trim($matchesFix[5], '.'));
-                    $ver2 = str_Replace(['x-dev', 'x'], ['0', '0'], trim($matchesFix[12], '.'));
+                    $ver1 = str_Replace(['-dev', 'x'], ['', '0'], trim($matchesFix[5], '.'));
+                    $ver2 = str_Replace(['-dev', 'x'], ['', '0'], trim($matchesFix[12], '.'));
                     // If the version is short, add enough '.0' to correct it.
                     while (substr_count($ver1, '.') < 2) {
                         $ver1 .= '.0';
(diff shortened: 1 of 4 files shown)


jonathan1055:
The test failure is one PHPstan problem that exists in the 8.x-3.x branch anyway. The code is clean using phpcs and the modified tests pass

coder_sniffer/Drupal/Sniffs/Commenting/DeprecatedSniff.php
```diff
@@ -75,20 +74,72 @@ public function process(File $phpcsFile, $stackPtr)
...
+                // Get just the first line of the text.
+                $key   = array_keys($textItems)[0];
+                $text1 = $textItems[$key];
+                preg_match('/^(.*)(as of|in) (drupal|)( |:|)+([\d\.\-xdev\?]+)(,| |. |)(.*)(removed|removal)([ |from|before|in|the]*) (drupal|)( |:|)([\d\-\.xdev]+)( |,|$)+(?:release|)(?:[\.,])*(.*)$/i', $text1, $matchesFix);
+
+                if (count($matchesFix) >= 12) {
+                    // This problem is a drupal core file and is fixable.
+                    if (empty($matchesFix[1]) === false) {
+                        // Verify that its acceptable to remove the text in [1].
+                        echo('>>> File: '.$phpcsFile->path.' line '.$tokens[($stackPtr)]['line'].PHP_EOL);
```
klausi: hm, I think we should not use echo() in Coder. All output should be done with addError() or addWarning().

Or this for debugging? Then we should hide this behind $this->debug check. See ScopeIndentSniff for example.

coder_sniffer/Drupal/Sniffs/Commenting/DeprecatedSniff.php
```diff
@@ -75,20 +74,72 @@ public function process(File $phpcsFile, $stackPtr)
...
+                    // If $correctedText is longer than 65 this will make the whole
+                    // line exceed 80. It is probably longer than 80 before the fix.
+                    if (strlen($correctedText) > 65) {
+                        echo('>>> WARNING: File '.$phpcsFile->path.' line '.$tokens[($stackPtr)]['line'].PHP_EOL);
+                        echo('>>> WARNING: $correctedText = '.$correctedText.PHP_EOL);
+                        echo('>>> WARNING: length of $correctedText = '.strlen($correctedText).' (so line will exceed 80)'.PHP_EOL);
+                    }
+
+                    $fix = $phpcsFile->addFixableError($error, $key, 'IncorrectTextLayoutFixable', [$fullText]);
+                    $phpcsFile->fixer->replaceToken($key, $correctedText);
```
klausi: only if $fix is true we are allowed to replace stuff. See other sniffs how they fix things.

coder_sniffer/Drupal/Sniffs/Commenting/DeprecatedSniff.php
```diff
@@ -75,20 +74,72 @@ public function process(File $phpcsFile, $stackPtr)
...
+                    $correctedText = trim('in drupal:'.$ver1.' and is removed from drupal:'.$ver2.'. '.trim($matchesFix[14]));
+                    // If $correctedText is longer than 65 this will make the whole
+                    // line exceed 80. It is probably longer than 80 before the fix.
+                    if (strlen($correctedText) > 65) {
+                        echo('>>> WARNING: File '.$phpcsFile->path.' line '.$tokens[($stackPtr)]['line'].PHP_EOL);
+                        echo('>>> WARNING: $correctedText = '.$correctedText.PHP_EOL);
+                        echo('>>> WARNING: length of $correctedText = '.strlen($correctedText).' (so line will exceed 80)'.PHP_EOL);
+                    }
+
+                    $fix = $phpcsFile->addFixableError($error, $key, 'IncorrectTextLayoutFixable', [$fullText]);
```
klausi: we should not have different error codes for fixable errors. Just use 1 code.

klausi (changes requested):
We need a DeprecatedUnitTest.inc.fixed test file where we demonstrate all the fixes that work.

coder_sniffer/Drupal/Sniffs/Commenting/DeprecatedSniff.php
```diff
@@ -75,20 +74,72 @@ public function process(File $phpcsFile, $stackPtr)
...
+            // All of the standard text should be on the first comment line, so
+            // try to match with common formatting errors to allow an automatic
+            // fix. If not possible then report a normal error.
+            $matchesFix = [];
+            $fix        = null;
+            if (count($textItems) > 0) {
+                // Get just the first line of the text.
+                $key   = array_keys($textItems)[0];
+                $text1 = $textItems[$key];
+                preg_match('/^(.*)(as of|in) (drupal|)( |:|)+([\d\.\-xdev\?]+)(,| |. |)(.*)(removed|removal)([ |from|before|in|the]*) (drupal|)( |:|)([\d\-\.xdev]+)( |,|$)+(?:release|)(?:[\.,])*(.*)$/i', $text1, $matchesFix);
```
klausi, line 127: we should not hard-code "drupal" here, right?

jonathan1055: Thanks for all the review comments. Very helpful. I will start with this one, as it affects how I proceed with the others. If we do not hard-code 'drupal' here then it would mean catering for contrib within the fixable section. I guess that preg_match could be altered to allow any word+undercore instead of 'drupal' but that would mean later testing to see if it was 'drupal' or blank (in which case the fix is for core) or some other word (in which case the fix is for contrib). The version checking would have to be different for contrib, but that could be done. 

If we think this is worth it, I will try to make the changes. But I'm not sure how much deprecation there is in contrib, so I was happy initially to exclude it from automatic fixing. 

The hard-coding of 'drupal' at that line effectively says "I am only attempting to provide auto-fixes for core problems". If I made that clearer in the comments would that be an acceptable solution? I think the alternative (if you do not want 'drupal' hard-coded there) would be for  me to re-write the whole patch to remove the auto-fixable part completely. But it does need the new sniff to cater for missing 'extra info' so there would be some work to do.

klausi: Yep, hard coding "drupal" just for the fixable error as first step would be acceptable I think. Please add code comments to explain that.

Thank you!

coder_sniffer/Drupal/Sniffs/Commenting/DeprecatedSniff.php
jonathan1055: It was for debug, and useful to keep in, so I will use your hint about $this->debug. Is it ok to call the runtime parameter `deprecated_debug` following the similarly named `scope_indent_debug` ?

coder_sniffer/Drupal/Sniffs/Commenting/DeprecatedSniff.php
jonathan1055: OK, will do

coder_sniffer/Drupal/Sniffs/Commenting/DeprecatedSniff.php
jonathan1055: OK

jonathan1055:
I have added a new DeprecatedUnitTest.inc.fixed test to verify the auto-fixes, improved the comments, added a debug parameter and fixed the actual code for coding standards. My travis test runs green, with no phpstan or phpcs warnings either.

Unfortunately I had merged in from 8.x-3.x the other recent commits, so they are also showing above. My new changes are all squashed down into one commit, the last one shown above. Is that acceptable for you to work with?

coder_sniffer/Drupal/Sniffs/Commenting/DeprecatedSniff.php
```diff
@@ -198,19 +198,24 @@ public function process(File $phpcsFile, $stackPtr)
...
-            // sniff FunctionCommentSniff.SeePunctuation so allow that sniff to
-            // report it (and fix it).
-        } else if (empty($matches) === true) {
-            $error = "The @see url '%s' does not match the standard: http(s)://www.drupal.org/node/n or http(s)://www.drupal.org/project/aaa/issues/n";
-            $phpcsFile->addWarning($error, $seeTag, 'DeprecatedWrongSeeUrlFormat', [$crLink]);
+        // If the @see tag exists but has no content then do nothing more as
+        // this is covered by FunctionComment.EmptySees sniff. It also causes
+        // an odd message with '<?php' shown for the @see url content.
+        if (!empty($string)) {
+            // Check the format of the @see url.
```
jonathan1055: This is just a simple change, wrapping the url check in a conditional, to avoid doing anything when there is nothing to check. This is covered by FunctionComment.EmptySees anyway.

jonathan1055: Reverted this change, as the other sniff is only for functions, so was not always being shown. Instead cater for blank @see url by setting $crLink to blank. Two commits squashed together for ease of review.

README.md
```diff
@@ -129,7 +129,7 @@ Editors:
 - [Visual Studio Code (VSCode)](https://www.drupal.org/node/1419996)
 
 
-## Automated Testing (PHPUnit + PHPCS)
+## Automated Testing (PHPUnit + PHPCS + PHPStan)
```
klausi: Unrelated change - looks like your merged  in the 8.x-3.x branch wrong?

klausi (changes requested):
Can you merge in the 8-x.3.x branch correctly so that we see only your changes here?

README.md
jonathan1055: I have now pushed a nice clean branch. Only the four files changed. I have left it as two commits, the first is the same a patch 16 which you reviewed. So you can see the changes I made following review.

coder_sniffer/Drupal/Test/Commenting/DeprecatedUnitTest.inc
```diff
@@ -2,9 +2,9 @@
 
 /**
  * @file
+ * This file contains test cases for DeprecatedSniff.
```
jonathan1055, line 5: These changes are to make the test file clean for all coding standards (but I think on reflection this was unnecessary as the test directories are ignored. Oh well ...)

coder_sniffer/Drupal/Sniffs/Commenting/DeprecatedSniff.php
```diff
@@ -114,15 +200,22 @@ public function process(File $phpcsFile, $stackPtr)
...
-        // If matches[4] is not blank it means that the url is correct but it
-        // ends with a period. As this can be a common mistake give a specific
-        // message to assist in fixing.
         if (isset($matches[4]) === true && empty($matches[4]) === false) {
-            $error = "The @see url '%s' should not end with a period.";
-            $phpcsFile->addWarning($error, $seeTag, 'DeprecatedPeriodAfterSeeUrl', [$crLink]);
+            // If matches[4] is not blank it means that the url is correct but
+            // it ends with a period. This is covered by the generic fixable
+            // sniff FunctionCommentSniff.SeePunctuation so allow that sniff to
+            // report it (and fix it).
```
klausi, line 218: This is an empty if() statement now and should be removed?

klausi (approved):
Cool, looks good from a visual review. One minor thing I would change, but I think that should not hold up progress here and can be fixed later.

jonathan1055:
Thanks for commiting. I can make a new PR for removing the empty `if` statement.

If I had known that the new tag 8.x-3.8 was imminent I would have worked on this a few days earlier to get it in, so that Core can start using the sniffs. Do you have a plan for when 3.9 will be released?

klausi:
Usually I make releases every 3 months. but if you need to get a release out earlier for drupal core I think we can do it anytime. Best you do a little bit of testing in the next days / weeks and then we could do a new release.

jonathan1055:
Thanks for the info. I may try to work on the trigger_error auto correction too, and then they can both get into the next release. Those two sniffs are closely related.

PR #95, opened 2020-03-07
docs(README): Add PHPStan command

arkener:
https://www.drupal.org/project/coder/issues/3118572

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ Editors:
 - [Visual Studio Code (VSCode)](https://www.drupal.org/node/1419996)
 
 
-## Automated Testing (PHPUnit + PHPCS)
+## Automated Testing (PHPUnit + PHPCS + PHPStan)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -144,6 +144,10 @@ Then execute the coding standards checker on Coder itself:
 
     ./vendor/bin/phpcs
 
+Then execute static analysis:
+
+    ./vendor/bin/phpstan
```
klausi: the correct command is `./vendor/bin/phpstan analyze`

klausi (changes requested):
Thanks, makes sense! Just one small thing. Can you also open a quick issue on drupal.org so that we can assign credits?

README.md
arkener: Thanks!, the PR has been updated with the proper command. I've used the `analyse` version, as the `analyze` command is the alias for the `analyse` command.

PR #100, opened 2020-04-11
Add a sniff to fix class name or interface name not identical with the filename

ranqiangjun:
https://www.drupal.org/project/coder/issues/3126781


ranqiangjun:
> Yes, found it, it's Drupal.Classes.ClassFileName, Thank you @klausi!

PR #101, opened 2020-04-11
Add a sniff to fix duplicate full stops in comments

ranqiangjun:
https://www.drupal.org/project/coder/issues/3126860

### DIFF
--- a/coder_sniffer/Drupal/Sniffs/Commenting/DocCommentSniff.php
+++ b/coder_sniffer/Drupal/Sniffs/Commenting/DocCommentSniff.php
@@ -248,6 +248,18 @@ class DocCommentSniff implements Sniff
             }
         }
 
+        if (strlen($shortContent) > 2 && substr($shortContent, -2) === '..') {
+            $theLast3rdChar = substr($shortContent, -3, 1);
+            // Ignore ... and /.. .
+            if (in_array($theLast3rdChar, ['.', '/']) === false) {
+                $fix = $phpcsFile->addFixableError('Duplicate full-stops found', $stackPtr, 'DuplicateFullStop');
+                if ($fix === true) {
+                    $newContent = substr($shortContent, 0, -1);
+                    $phpcsFile->fixer->replaceToken($short, $newContent);
+                }
+            }
+        }
+
         if ($tokens[$short]['line'] !== $tokens[$shortEnd]['line']) {
             $error = 'Doc comment short description must be on a single line, further text should be a separate paragraph';
             $phpcsFile->addError($error, $shortEnd, 'ShortSingleLine');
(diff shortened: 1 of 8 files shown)


coder_sniffer/Drupal/Sniffs/Commenting/DuplicateFullStopSniff.php
```diff
@@ -0,0 +1,80 @@
...
+     */
+    public function process(File $phpcsFile, $stackPtr)
+    {
+        $tokens         = $phpcsFile->getTokens();
+        $content        = $tokens[$stackPtr]['content'];
+        $contentTrimmed = trim($content);
+
+        if (substr($contentTrimmed, -2) === '..' && substr($contentTrimmed, -3) !== '...') {
+            $phpcsFile->addFixableWarning('Duplicate full-stops found', $stackPtr, 'DuplicateFullStop');
+            if ($phpcsFile->fixer !== null) {
```
klausi: please use the return value of addFixableWarning() to check if the fixer is enabled like we do in other sniffs.

coder_sniffer/Drupal/Sniffs/Commenting/DuplicateFullStopSniff.php
```diff
@@ -0,0 +1,80 @@
...
+                $contentReversed = strrev($content);
```
klausi: why this complicated fix? Can we use strrpos() for ".." and then assemble the replacement with substr() and only add one dot?

klausi (commented):
I think this sniff is not fully sufficient since it does not take comment paragraphs into account. Example:

```
// This is an example comment where I will demonstrate the usage of path ./../..
// in this sentence.
```

So I think we need to integrate this into our existing comment sniffs that check if a comment ends with a dot. DocCommentSniff, FunctionCommentSniff and InlineCommentSniff I think.

klausi (changes requested):
Requesting changes per comment above.

ranqiangjun:
Hi @klausi,  I will work on it in hours, thanks for your review!

tests/Drupal/Commenting/InlineCommentUnitTest.inc.fixed
```diff
@@ -112,3 +112,13 @@ function test2() {
 
   return $x;
 }
+
+// An example comment where I will demonstrate the usage of path ./../..
+// in this sentence.
```
ranqiangjun, line 117: Addressed the example.

coder_sniffer/Drupal/Sniffs/Commenting/DocCommentSniff.php
```diff
@@ -248,6 +248,18 @@ public function process(File $phpcsFile, $stackPtr)
             }
         }
 
+        if (strlen($shortContent) > 2 && substr($shortContent, -2) === '..') {
+            $theLast3rdChar = substr($shortContent, -3, 1);
+            // Ignore ... and /.. .
```
ranqiangjun, line 253: Ignore comments end with three dots and comments look like a path. `/..`

ranqiangjun (commented):
Self-review

coder_sniffer/Drupal/Sniffs/Commenting/DuplicateFullStopSniff.php
ranqiangjun: Did in this iteration.

coder_sniffer/Drupal/Sniffs/Commenting/DuplicateFullStopSniff.php
ranqiangjun: Here is for ignoring comments end with `...`, simplified a little in this/the new iteration

ranqiangjun:
It's an incomplete one, will continue when I get time. Only touched DocCommentSniff and InlineCommentSniff so far.

klausi:
Closing this pull request due to inactivity, please open a new one if you are working on this again!

PR #102, opened 2020-04-30
Make it possible to deprecate and point to semantic versions of contrib projects

eiriksm:
Now that we have semantic versioning, we should allow deprecation messages to point to these versions as well.

We might simplify the logic here now and remove the drupal:n.n.n altogether, but that would be part of the discussion here I guess? :)

### DIFF
--- a/coder_sniffer/Drupal/Sniffs/Commenting/DeprecatedSniff.php
+++ b/coder_sniffer/Drupal/Sniffs/Commenting/DeprecatedSniff.php
@@ -169,14 +169,17 @@ class DeprecatedSniff implements Sniff
             }
         } else {
             // The text follows the basic layout. Now check that the versions
-            // match drupal:n.n.n or project:n.x-n.n or project:n.x-n.n-version[n].
+            // match drupal:n.n.n or project:n.x-n.n or project:n.x-n.n-version[n]
+            // or project:n.n.n or project:n.n.n-version[n].
             // The text must be all lower case and numbers can be one or two digits.
             foreach (['deprecation-version' => $matches[1], 'removal-version' => $matches[2]] as $name => $version) {
                 if (preg_match('/^drupal:\d{1,2}\.\d{1,2}\.\d{1,2}$/', $version) === 0
                     && preg_match('/^[a-z\d_]+:\d{1,2}\.x\-\d{1,2}\.\d{1,2}$/', $version) === 0
                     && preg_match('/^[a-z\d_]+:\d{1,2}\.x\-\d{1,2}\.\d{1,2}-[a-z]{1,5}\d{1,2}$/', $version) === 0
+                    && preg_match('/^[a-z\d_]+:\d{1,2}\.\d{1,2}\.\d{1,2}$/', $version) === 0
+                    && preg_match('/^[a-z\d_]+:\d{1,2}\.\d{1,2}\.\d{1,2}-[a-z]{1,5}\d{1,2}$/', $version) === 0
                 ) {
-                    $error = "The %s '%s' does not match the lower-case machine-name standard: drupal:n.n.n or project:n.x-n.n or project:n.x-n.n-version[n]";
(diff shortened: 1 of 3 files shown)


klausi:
Thanks, can you make a drupal.org Coder issue for this? That way we can track status and credits there.

We should also get feedback from @jonathan1055 who wrote this sniff.

Can you also merge in the 8.x-3.x branch? Then phpstan should not fail anymore on Travis CI.

klausi:
Please also add an example to the test cases so that we verify the behavior there as well.

eiriksm:
Added an issue here: https://www.drupal.org/project/coder/issues/3138133

eiriksm:
...and a test. Should I have made that comment here or there?

coder_sniffer/Drupal/Sniffs/Commenting/DeprecatedSniff.php
```diff
@@ -169,14 +169,17 @@ public function process(File $phpcsFile, $stackPtr)
...
-            // match drupal:n.n.n or project:n.x-n.n or project:n.x-n.n-version[n].
+            // match drupal:n.n.n or project:n.x-n.n or project:n.x-n.n-version[n]
+            // or project:n.n.n or project:n.n.n-version[n].
             // The text must be all lower case and numbers can be one or two digits.
             foreach (['deprecation-version' => $matches[1], 'removal-version' => $matches[2]] as $name => $version) {
                 if (preg_match('/^drupal:\d{1,2}\.\d{1,2}\.\d{1,2}$/', $version) === 0
                     && preg_match('/^[a-z\d_]+:\d{1,2}\.x\-\d{1,2}\.\d{1,2}$/', $version) === 0
                     && preg_match('/^[a-z\d_]+:\d{1,2}\.x\-\d{1,2}\.\d{1,2}-[a-z]{1,5}\d{1,2}$/', $version) === 0
+                    && preg_match('/^[a-z\d_]+:\d{1,2}\.\d{1,2}\.\d{1,2}$/', $version) === 0
+                    && preg_match('/^[a-z\d_]+:\d{1,2}\.\d{1,2}\.\d{1,2}-[a-z]{1,5}\d{1,2}$/', $version) === 0
```
klausi: I think we could collapse those 5 regexes now into just one. Something like this should work:
```php
preg_match('/^[a-z\d_]+:(\d{1,2}\.\d{1,2}\.\d{1,2}|\d{1,2}\.x\-\d{1,2}\.\d{1,2})(-[a-z]{1,5}\d{1,2})?$/', $version)
```

jonathan1055:
> and remove the drupal:n.n.n altogether

That's interesting, because if we allow projects to have n.n.n then yes we can remove the specific 'drupal' check, but this means we would not catch any typo for example `drpal:8.8.0` would be assumed to be a contrib project and would pass.

I know that @pfrenssen and @klausi already know this, but any extensions or additions to allowed standards I believe should be discussed and agreed on the [Coding Standards board](https://www.drupal.org/project/coding_standards), not just the Coder issue which implements it.

klausi:
I think we don't need to discuss in the Coding Standards issue queue since semantic version numbers for contrib are already a reality on drupal.org, see https://www.drupal.org/node/1015226#semver-transition

Do we need to update any other documentation?

jonathan1055:
ok.
Yes you'd need to update https://www.drupal.org/core/deprecation/#how-definitions 

Is there any way we detect typos such as `drpal:8.8.0`. Is there some clever regex which takes the string `drupal` and  matches anagrams with it, or strings of 5 letters where one has been missed?

klausi (changes requested):
We could check for single letter typos for Drupal core, yes.

I think that is out of scope for this issue, let's get semantic version support for contrib and a simplified regex done here first and then we can think of further checks in a new issue.

eiriksm:
Sorry for the delay. Collapsed all of the regexes into one now :hot_pepper:

PR #103, opened 2020-05-04
fix(CoreVersionRequirement): Ensure sniff is only run on project info.yml files

arkener:
Reference: https://www.drupal.org/project/coder/issues/3133361


arkener:
Tests are failing because of an upstream change https://github.com/pfrenssen/coder/pull/104

PR #104, opened 2020-05-04
tests(phpstan): Remove ignored false positives that have been fixed in PHPCS

arkener:
This issue has been fixed upstream in PHPCS via commit https://github.com/squizlabs/PHP_CodeSniffer/commit/51d44091ba9ba3fa7bdf8046dcc1c6768d0cc774


klausi:
Thank you so much!

PR #107, opened 2020-05-16
fix(composer): Prevent composer autoload warnings by moving the tests outside the autoload directory

klausi:
Issue: https://www.drupal.org/project/coder/issues/3116414


arkener:
Everything seems to be working fine, running the unit tests against the new codebase results in the same results as the current one.
```
PHPUnit 7.5.20 by Sebastian Bergmann and contributors.

................................................................. 65 / 96 ( 67%)
...............................                                   96 / 96 (100%)

Time: 5.77 seconds, Memory: 30.00 MB

OK (96 tests, 14 assertions)
```
Also checked if there were any references left to the old "Test" directories. The last references were removed in the latest commit, so everything seem good to go. Thanks for your hard work!

PR #108, opened 2020-05-18
feat(GlobalClass): Add FQN based class check and increase list of disallowed classes

arkener:
Reference: https://www.drupal.org/project/coder/issues/3137530

### DIFF
--- a/coder_sniffer/DrupalPractice/Sniffs/Objects/GlobalClassSniff.php
+++ b/coder_sniffer/DrupalPractice/Sniffs/Objects/GlobalClassSniff.php
@@ -25,18 +25,66 @@ class GlobalClassSniff implements Sniff
 {
 
     /**
-     * Class names that should not be called statically, mostly entity classes.
+     * Core class names that should not be called statically, mostly entity
+     * classes.
      *
      * @var string[]
      */
-    protected $classes = [
-        'File',
-        'Node',
-        'NodeType',
-        'Role',
-        'Term',
-        'User',
+    protected $coreClasses = [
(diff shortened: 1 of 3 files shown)


coder_sniffer/DrupalPractice/Sniffs/Objects/GlobalClassSniff.php
```diff
@@ -112,4 +169,53 @@ public function process(File $phpcsFile, $stackPtr)
...
 
+    /**
+     * Retrieve the fully qualified name of the given classname.
+     *
+     * @param \PHP_CodeSniffer\Files\File $phpcsFile The file being scanned.
+     * @param string                      $className The classname for which to retrieve the FQN.
+     *
+     * @return string
+     */
+    protected function getFullyQualifiedName(File $phpcsFile, $className)
```
klausi, line 180: this is a bit sad that we need to implement this ourselves. PHPCS core should have a helper for this. I don't see a better way currently, so I think this ok.

klausi (approved):
Looks good, thanks!

Leaving this for you to merge and to push to drupal.org :)

coder_sniffer/DrupalPractice/Sniffs/Objects/GlobalClassSniff.php
arkener: Yes, it would be nice to have an implementation for this upstream, for now this seems to be ok. If we ever need this information again then we might consider moving this to a coder util class or something :slightly_smiling_face:

PR #110, opened 2020-06-03
Add UseGlobalClassSniff.

cburschka:
Adds a sniff for global class imports. https://www.drupal.org/project/coder/issues/3120001


cburschka:
Fixed the bug that removed function imports, and a quirk that left blank lines between use statements (unlike extra blank lines before and after the block, these are currently not removed by other sniffs). Unit test expanded to cover that.

coder_sniffer/Drupal/Sniffs/Classes/UseGlobalClassSniff.php
```diff
@@ -0,0 +1,133 @@
...
+            $class     = $phpcsFile->findNext(T_STRING, $lineStart, $lineEnd);
+            $className = $tokens[$class]['content'];
+            // If there is more than one string token, the last one is the alias.
+            $alias     = $phpcsFile->findPrevious(T_STRING, $lineEnd, $stackPtr);
+            $aliasName = $tokens[$alias]['content'];
+
+            $error = 'Non-namespaced classes/interfaces/traits should not be referenced with use statements';
+            $phpcsFile->addFixableError($error, $class, 'RedundantUseStatement');
+
+            $phpcsFile->fixer->beginChangeset();
```
klausi: the fixer should only run if the return value of addFixableError() is true. You can check that in other classes that implement a fixer.

tests/Drupal/Classes/UseGlobalClassUnitTest.php
```diff
@@ -0,0 +1,67 @@
...
+
+
+    /**
+     * Returns a list of test files that should be checked.
+     *
+     * @param string $testFileBase The base path that the unit tests files will have.
+     *
+     * @return array<string>
+     */
+    protected function getTestFiles($testFileBase): array
```
klausi: this method can be removed, the default implementation will pick up the test file by the same name

klausi (changes requested):
Nice work, thank you!

I just have 2 minor comments, then I think we can merge this.

I briefly researched if there is already a sniff in upstream PHPCS but could not find anything. So I think it is ok that we invent a sniff here for now that really the whole PHP community would want otherwise.

coder_sniffer/Drupal/Sniffs/Classes/UseGlobalClassSniff.php
cburschka: Thanks! I'd only checked the FullyQualifiedNamespace sniff, and assumed the condition there was just for checking whether the error was fixable (which it always is here), not that the fixer is enabled as well.

PR #112, opened 2020-06-10
fix(OptionsTSniff): Fix false positive '#options' property have to run through t()

T2L:
fix(OptionsTSniff): Fix false positive '#options' property have to run through t() on non form element (#3062766 by zarabatana, Taran2L)

### DIFF
--- a/coder_sniffer/DrupalPractice/Sniffs/General/OptionsTSniff.php
+++ b/coder_sniffer/DrupalPractice/Sniffs/General/OptionsTSniff.php
@@ -84,15 +84,22 @@ class OptionsTSniff implements Sniff
         // We want to find if the element "#options" belongs to a form element.
         // Array with selectable options for a form element.
         $formElements = [
-          "'checkboxes'",
-          "'radios'",
-          "'select'",
-          "'tableselect'",
+            "'checkboxes'",
+            "'radios'",
+            "'select'",
+            "'tableselect'",
         ];
         // Find beggining of the array containing "#options" element.
-        $startArray = $phpcsFile->findPrevious(T_EQUAL, $stackPtr, null, false, "=");
+        $startArray = $phpcsFile->findPrevious(T_VARIABLE, $stackPtr);
+
         // Find next element on array of "#type".
-        $findType = $phpcsFile->findNext(T_CONSTANT_ENCAPSED_STRING, ($startArray + 1), null, false, "'#type'");
+        $findType = $phpcsFile->findNext(T_CONSTANT_ENCAPSED_STRING, ($startArray + 1), $statementEnd, false, "'#type'");
+
+        // Stop checking the array if its #type cannot be determined.
+        if ($findType === false) {
+            return;
+        }
+
         // Get the value of "#type".
         $valueType = $phpcsFile->findNext(T_CONSTANT_ENCAPSED_STRING, ($findType + 1), null, false);
 
--- a/coder_sniffer/DrupalPractice/Sniffs/General/OptionsTSniff.php
+++ b/coder_sniffer/DrupalPractice/Sniffs/General/OptionsTSniff.php
@@ -89,8 +89,8 @@ class OptionsTSniff implements Sniff
             "'select'",
             "'tableselect'",
         ];
-        // Find beggining of the array containing "#options" element.
-        $startArray = $phpcsFile->findPrevious(T_VARIABLE, $stackPtr);
+        // Find beginning of the array containing "#options" element.
+        $startArray = $phpcsFile->findStartOfStatement($stackPtr, [T_DOUBLE_ARROW, T_OPEN_SHORT_ARRAY, T_OPEN_PARENTHESIS, T_COMMA]);
 
         // Find next element on array of "#type".
         $findType = $phpcsFile->findNext(T_CONSTANT_ENCAPSED_STRING, ($startArray + 1), $statementEnd, false, "'#type'");
--- a/coder_sniffer/DrupalPractice/Sniffs/General/OptionsTSniff.php
+++ b/coder_sniffer/DrupalPractice/Sniffs/General/OptionsTSniff.php
@@ -81,6 +81,21 @@ class OptionsTSniff implements Sniff
             $statementEnd = $tokens[$arrayToken]['bracket_closer'];
         }
 
+        // We want to find if the element "#options" belongs to a form element.
+        // Array with selectable options for a form element.
+        $formElements = [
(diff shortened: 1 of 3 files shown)


T2L:
Huh, tests fail on:

```
--------------------------------------------------------------------------------
FOUND 4 ERRORS AFFECTING 4 LINES
--------------------------------------------------------------------------------
 87 | ERROR | [x] Array key not indented correctly; expected 12 spaces but
    |       |     found 10
 88 | ERROR | [x] Array key not indented correctly; expected 12 spaces but
    |       |     found 10
 89 | ERROR | [x] Array key not indented correctly; expected 12 spaces but
    |       |     found 10
 90 | ERROR | [x] Array key not indented correctly; expected 12 spaces but
    |       |     found 10
--------------------------------------------------------------------------------
PHPCBF CAN FIX THE 4 MARKED SNIFF VIOLATIONS AUTOMATICALLY
--------------------------------------------------------------------------------
```

but the indentation is correct, I guess we need another issue :)

coder_sniffer/DrupalPractice/Sniffs/General/OptionsTSniff.php
```diff
@@ -81,6 +81,21 @@ public function process(File $phpcsFile, $stackPtr)
...
         }
 
+        // We want to find if the element "#options" belongs to a form element.
+        // Array with selectable options for a form element.
+        $formElements = [
+          "'checkboxes'",
+          "'radios'",
+          "'select'",
+          "'tableselect'",
+        ];
```
arkener, line 91: Coder uses 4 space based indenting, this array currently uses 2 space based indenting.

coder_sniffer/DrupalPractice/Sniffs/General/OptionsTSniff.php
```diff
@@ -81,6 +81,21 @@ public function process(File $phpcsFile, $stackPtr)
...
+        // Find beggining of the array containing "#options" element.
+        $startArray = $phpcsFile->findPrevious(T_EQUAL, $stackPtr, null, false, "=");
```
arkener: Single quotes can be used here for the '=' value.

We should add an $end value here to scope the array definition, as this search will currently keep searching until it finds a '=', which could result in finding a different form element. For example, the following code.
```
$form['lorem'] = [
  '#type' => 'checkboxes',
  '#options' => [
    'lorem' => 'Lorem',
  ],
];

$form['ipsum']['#options'] = [
  'attributes' => [
    'target' => '_blank',
  ],
];
```
Will result in:
```
--------------------------------------------------------------------------------
  4 | WARNING | #options values usually have to run through t() for translation
 10 | WARNING | #options values usually have to run through t() for translation
--------------------------------------------------------------------------------
```

coder_sniffer/DrupalPractice/Sniffs/General/OptionsTSniff.php
```diff
@@ -81,6 +81,21 @@ public function process(File $phpcsFile, $stackPtr)
...
+        // Find next element on array of "#type".
+        $findType = $phpcsFile->findNext(T_CONSTANT_ENCAPSED_STRING, ($startArray + 1), null, false, "'#type'");
+        // Get the value of "#type".
+        $valueType = $phpcsFile->findNext(T_CONSTANT_ENCAPSED_STRING, ($findType + 1), null, false);
```
arkener, line 104: $startArray, $findType and $valueType should check if the "find" results in a false, if this is the case then we can just stop checking this array as we can't determine the form element type.  Currently it will just calculate `false + 1 = 1`, after which it will start searching at the top of the document.

arkener (changes requested):
Thank you for the changes, I've left some comments on your PR.

coder_sniffer/DrupalPractice/Sniffs/General/OptionsTSniff.php
T2L: Oops, missed it. My bad

coder_sniffer/DrupalPractice/Sniffs/General/OptionsTSniff.php
T2L: I assume we can look for a previous variable here?

coder_sniffer/DrupalPractice/Sniffs/General/OptionsTSniff.php
```diff
@@ -81,6 +81,28 @@ public function process(File $phpcsFile, $stackPtr)
...
+        // We want to find if the element "#options" belongs to a form element.
+        // Array with selectable options for a form element.
+        $formElements = [
+            "'checkboxes'",
+            "'radios'",
+            "'select'",
+            "'tableselect'",
+        ];
+        // Find beggining of the array containing "#options" element.
+        $startArray = $phpcsFile->findPrevious(T_VARIABLE, $stackPtr);
```
arkener: It's probably better better to implement findStartOfStatement() here, to ensure we don't catch an unexpected variable here like the following, which would check the $label:
```
$form['lorem'] = [
  '#type' => 'checkboxes',
  '#title' => $label,
  '#options' => [
    'lorem' => 'Lorem',
  ],
];
```

We should be able to implement it like:
`$startArray = $phpcsFile->findStartOfStatement($stackPtr, [T_DOUBLE_ARROW, T_OPEN_SHORT_ARRAY, T_OPEN_PARENTHESIS, T_COMMA]);`

coder_sniffer/DrupalPractice/Sniffs/General/OptionsTSniff.php
```diff
@@ -81,6 +81,28 @@ public function process(File $phpcsFile, $stackPtr)
...
 
+        // We want to find if the element "#options" belongs to a form element.
+        // Array with selectable options for a form element.
+        $formElements = [
+            "'checkboxes'",
+            "'radios'",
+            "'select'",
+            "'tableselect'",
+        ];
+        // Find beggining of the array containing "#options" element.
```
arkener: Minor nitpicking, but there is a typo in "beggining" :relaxed:

arkener (changes requested):
Some minor changes left and then we're good to go.

coder_sniffer/DrupalPractice/Sniffs/General/OptionsTSniff.php
T2L: Thanks for the example, it is super helpful!

arkener (approved):
Looks great, thanks!

PR #113, opened 2020-06-30
Issue #3155865 add ddm() to discouraged function sniff

jonathan1055:
Drupal issue https://www.drupal.org/project/coder/issues/3155865
Function `dd()` is already discouraged but this is now named `ddm()` so that should also be in the discouraged list.


jonathan1055:
I noticed that only two of the 23 discouraged debug functions have test coverage:
````
/**
 * Debugging functions are discouraged.
 */
function test27() {
  $var = array(1, 2, 3);
  dsm($var);
  dpm($var);
}
````
If only a sample of the functions are tested it seems we do not need to cover every one of them. So I did not make any change to the test.

jonathan1055:
Just noticed that I added `ddm` under `dd` but if you want strict alphabetical order it should be one line further down, after `ddebug_backtrace`

arkener:
Not updating the test case for this rule is fine, as the test is mostly there to check if the rule itself works correctly. As you've already stated, I would prefer if the function list is sorted alphabetically (I know, nitpicking :relaxed:)

jonathan1055:
Done! I was not sure if you accept forced pushes here, so did not squash the commits. You can do that on merge, I am sure.

PR #114, opened 2020-06-30
Issue #3153448 line length should not be ignored for 2-element arrays

jonathan1055:
D.O Issue https://www.drupal.org/project/coder/issues/3153448
Line length limit should not be ignored for 2-element arrays
This first commit just changed the test data to demonstrate this. The tests should fail to detect the error.

### DIFF
--- a/coder_sniffer/Drupal/Sniffs/Arrays/ArraySniff.php
+++ b/coder_sniffer/Drupal/Sniffs/Arrays/ArraySniff.php
@@ -31,6 +31,8 @@ class ArraySniff implements Sniff
     /**
      * The limit that the length of a line should not exceed.
      *
+     * This can be configured to have a different value but the default is 80.
+     *
      * @var integer
      */
     public $lineLimit = 80;
@@ -109,17 +111,14 @@ class ArraySniff implements Sniff
         }
 
         if ($isInlineArray === true) {
-            // Check if this array contains at least 3 elements and exceeds the 80
-            // character line length.
+            // Check if this array has more than one element and exceeds the
+            // line length defined by $this->lineLimit.
             if ($tokens[$tokens[$stackPtr][$parenthesisCloser]]['column'] > $this->lineLimit) {
                 $comma1 = $phpcsFile->findNext(T_COMMA, ($stackPtr + 1), $tokens[$stackPtr][$parenthesisCloser]);
                 if ($comma1 !== false) {
-                    $comma2 = $phpcsFile->findNext(T_COMMA, ($comma1 + 1), $tokens[$stackPtr][$parenthesisCloser]);
-                    if ($comma2 !== false) {
-                        $error = 'If the line declaring an array spans longer than %s characters, each element should be broken into its own line';
-                        $data  = [$this->lineLimit];
-                        $phpcsFile->addError($error, $stackPtr, 'LongLineDeclaration', $data);
-                    }
+                    $error = 'If the line declaring an array spans longer than %s characters, each element should be broken into its own line';
+                    $data  = [$this->lineLimit];
+                    $phpcsFile->addError($error, $stackPtr, 'LongLineDeclaration', $data);
                 }
             }
 
--- a/coder_sniffer/Drupal/Sniffs/Arrays/ArraySniff.php
+++ b/coder_sniffer/Drupal/Sniffs/Arrays/ArraySniff.php
@@ -114,9 +114,9 @@ class ArraySniff implements Sniff
             // Check if this array has more than one element and exceeds the
             // line length defined by $this->lineLimit.
             $currentLine = $tokens[$stackPtr]['line'];
-            $tokenCount = $stackPtr;
-            while ($tokenCount < $phpcsFile->numTokens-1 && $tokens[$tokenCount+1]['line'] === $currentLine) {
-               $tokenCount++;
+            $tokenCount  = $stackPtr;
+            while ($tokenCount < ($phpcsFile->numTokens - 1) && $tokens[($tokenCount + 1)]['line'] === $currentLine) {
+                $tokenCount++;
             };
             $lineLength = ($tokens[$tokenCount]['column'] + $tokens[$tokenCount]['length'] - 1);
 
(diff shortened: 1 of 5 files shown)


jonathan1055:
As expected we get test fails
````
There was 1 failure:
1) Drupal\Test\Arrays\ArrayUnitTest::testSniff
[LINE 13] Expected 1 error(s) in ArrayUnitTest.inc but found 0 error(s).
[LINE 33] Expected 1 error(s) in ArrayUnitTest.inc but found 0 error(s).
````

Those lines now have only two elements but extend beyond the limit.

**Comment by jonathan1055** [comment]
This actually needs a bit more work. The current sniff checks the column position of the closing parenthesis. If this is not on the same line as the multiple long elements then no warning is given at all. For example - the following is not reported and completely passes all coder checks:
````
$x = ['123', '456', 'long string which extends beyond the limit', 'another element',
]
````
Maybe the length check should get the position of the comma not the closing parenthesis?

**Comment by jonathan1055** [comment]
The last change gets the proper line length, not just using the position of the parenthesisCloser. There should always be one more character after the closer for an inline array (the semicolon) but it is not good to assume that. Better to get the actual line length (I re-used a function `LineLengthSniff::getLineLength`). 

The test will fail, showing that the current test data has a line of length 101 which was not flagged as an error before.

**Comment by jonathan1055** [comment]
As expected we get one failure
````
1) Drupal\Test\Arrays\ArrayUnitTest::testSniff
[LINE 15] Expected 0 error(s) in ArrayUnitTest.1.inc but found 1 error(s). The error(s) found were:
 -> The array declaration line has 101 characters (the limit is 100). Each element should be on a new line (Drupal.Arrays.Array.LongLineDeclaration)
````

File: tests/Drupal/Arrays/ArrayUnitTest.inc [hunk 1]
```diff
@@ -10,7 +10,7 @@ $array = [
   'inline' => [],
   'inline3' => ['one', 'two', 'three'],
   'inline_long_ok' => ['one', 'two', 'three', 'four', 'five'],
-  'inline_long_nok' => ['one', 'two', 'three', 'four', 'five', 'six', 'seven', 'eight'],
```
klausi: we should not change the existing test case, please add a new one.

coder_sniffer/Drupal/Sniffs/Arrays/ArraySniff.php
```diff
@@ -109,23 +111,26 @@ public function process(File $phpcsFile, $stackPtr)
...
+            if ($lineLength > $this->lineLimit) {
                 $comma1 = $phpcsFile->findNext(T_COMMA, ($stackPtr + 1), $tokens[$stackPtr][$parenthesisCloser]);
                 if ($comma1 !== false) {
-                    $comma2 = $phpcsFile->findNext(T_COMMA, ($comma1 + 1), $tokens[$stackPtr][$parenthesisCloser]);
-                    if ($comma2 !== false) {
-                        $error = 'If the line declaring an array spans longer than %s characters, each element should be broken into its own line';
-                        $data  = [$this->lineLimit];
-                        $phpcsFile->addError($error, $stackPtr, 'LongLineDeclaration', $data);
-                    }
+                    $error = 'The array declaration line has %s characters (the limit is %s). Each element should be on a new line';
```
klausi: I think there can be edge cases with arrays with just 1 item like
```php
$array = array(very_long_function_call_i_tell_you($parameter1, $parameter2, $parameter3, $parameter4, $parameter5));
```
But even such a declaration should be split up over multiple lines. Only our error message is a bit inaccurate because the array has only 1 element. We could use "The array declaration line has %s characters (the limit is %s). The array content should be split up over multiple lines"

coder_sniffer/Drupal/Sniffs/Arrays/ArraySniff.php
```diff
@@ -109,23 +111,26 @@ public function process(File $phpcsFile, $stackPtr)
...
         if ($isInlineArray === true) {
-            // Check if this array contains at least 3 elements and exceeds the 80
-            // character line length.
-            if ($tokens[$tokens[$stackPtr][$parenthesisCloser]]['column'] > $this->lineLimit) {
+            // Check if this array has more than one element and exceeds the
+            // line length defined by $this->lineLimit.
+            $currentLine = $tokens[$stackPtr]['line'];
+            $tokenCount  = $stackPtr;
+            while ($tokenCount < ($phpcsFile->numTokens - 1) && $tokens[($tokenCount + 1)]['line'] === $currentLine) {
+                $tokenCount++;
```
klausi, line 119: why do we need to count here? We already know the array is on a single line. We can use the column from $parenthesisCloser.

klausi (changes requested):
Thanks, a couple of things we should improve.

coder_sniffer/Drupal/Sniffs/Arrays/ArraySniff.php
jonathan1055: See https://github.com/pfrenssen/coder/pull/114#issuecomment-652974349 at 2 Jul 13:24 BST above. The `$parenthesisCloser` is not necessarily the last thing on the line, there will often be a comma or semi-colon, but there may be more and we cannot assume anything. So this count gets all the content on that line to make sure we get the accurate length.

tests/Drupal/Arrays/ArrayUnitTest.inc
jonathan1055: OK will do.

coder_sniffer/Drupal/Sniffs/Arrays/ArraySniff.php
jonathan1055: OK, will change the message.

tests/Drupal/Arrays/ArrayUnitTest.inc
jonathan1055: I have reverted the changed test lines, and added new instead. This also showed where a previous line had 101 chars but was allowed to pass, hence the change in getErrorList()

coder_sniffer/Drupal/Sniffs/Arrays/ArraySniff.php
jonathan1055: I've changed the message text as requested.

jonathan1055:
@klausi Ready for re-review

jonathan1055:
Thanks.

PR #115, opened 2020-08-03
feat(InlineVariableComment): Add sniff for @var inline variable type declarations

arkener:
Reference: https://www.drupal.org/project/coder/issues/3163106

### DIFF
--- a/coder_sniffer/Drupal/Sniffs/Commenting/InlineVariableCommentSniff.php
+++ b/coder_sniffer/Drupal/Sniffs/Commenting/InlineVariableCommentSniff.php
@@ -1,0 +1,131 @@
+<?php
+
+/**
+ * \Drupal\Sniffs\Commenting\InlineVariableCommentSniff
+ *
+ * @category PHP
+ * @package  PHP_CodeSniffer
+ * @link     http://pear.php.net/package/PHP_CodeSniffer
+ */
+
+namespace Drupal\Sniffs\Commenting;
+
+use PHP_CodeSniffer\Files\File;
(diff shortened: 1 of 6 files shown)


coder_sniffer/Drupal/Sniffs/Commenting/InlineVariableCommentSniff.php
```diff
@@ -0,0 +1,131 @@
...
+            T_VAR,
+        ];
+
+        // If this is a function/class/interface doc block comment, skip it.
+        $nextToken = $phpcsFile->findNext(Tokens::$emptyTokens, ($stackPtr + 1), null, true);
+        if (in_array($tokens[$nextToken]['code'], $ignore, true) === true) {
+            return;
+        }
+
+        if ($tokens[$stackPtr]['code'] !== T_DOC_COMMENT_TAG && strpos($tokens[$stackPtr]['content'], '/*') === 0) {
```
klausi: I would write `$tokens[$stackPtr]['code'] === T_COMMENT` to be more explicit that this is about `/*` style comments?

klausi (approved):
Nice work, thanks!

I just have a minor comment that you can fix or not, then ready to be merged!

coder_sniffer/Drupal/Sniffs/Commenting/InlineVariableCommentSniff.php
arkener: Good point, changing this to `T_COMMENT` would also catch `#` and `//` based comments which should also be changed to `/** */`. I've added an extra check for this and expanded the tests to includes these cases.

PR #116, opened 2020-08-03
feat(ClassComment): Add a sniff to check for "Class Example" doc comments

arkener:
Reference: https://www.drupal.org/project/coder/issues/3156856


tests/Drupal/Commenting/ClassCommentUnitTest.1.inc
```diff
@@ -0,0 +1,13 @@
+<?php
+
+/**
+ * @file
+ * Testing class/trait comments.
```
klausi: any reason we need to start a new file?

We could just expand ClassCommentUnitTest.inc?

klausi (approved):
Makes sense, thanks!

Can you move the test into the first file? I think we don't need to start a new one, the existing one is also quite short.

After that you can merge!

tests/Drupal/Commenting/ClassCommentUnitTest.1.inc
arkener: The ClassCommentUnitTest.inc currently tests the "Missing" part of the ClassComment rule. We could merge this test in this file but it would replace the existing test, as it can't be missing and too short at the same time.

klausi: I meant adding a second class declaration to the test file, like in Drupal 7 code. Anyway, I don't care too much - having a second test file is also fine.

klausi (approved):

tests/Drupal/Commenting/ClassCommentUnitTest.1.inc
arkener: Ah, yes that makes sense, I'll merge the test into the existing test file and then merge this. Thanks for the review!

PR #118, opened 2020-09-10
Create @todo format sniff.

adamzimmermann:
Adds @todo format sniff.

Drupal.org issue:
- https://www.drupal.org/project/coder/issues/2222435

Testing/Examples:
- https://regex101.com/r/cir1XP/9

Questions:
- Will the `// ` be in the string that is sent to the sniff? If so, I need to adjust it. 
- Should we enforce a capital letter after `@todo`? (I went ahead and did this.)

### DIFF
--- a/coder_sniffer/Drupal/Sniffs/Commenting/TodoCommentSniff.php
+++ b/coder_sniffer/Drupal/Sniffs/Commenting/TodoCommentSniff.php
@@ -49,11 +49,14 @@ class TodoCommentSniff implements Sniff
      */
     public function process(File $phpcsFile, $stackPtr)
     {
-        $tokens = $phpcsFile->getTokens();
+        $tokens  = $phpcsFile->getTokens();
+        // Use the comment text by default.
         $comment = $tokens[$stackPtr]['content'];
-        if ($tokens[$stackPtr]['code'] == T_DOC_COMMENT_TAG) {
+        // Use the next line for multi-line comments.
+        if ($tokens[$stackPtr]['code'] === T_DOC_COMMENT_TAG) {
             $comment = $phpcsFile->findNext(T_DOC_COMMENT_STRING, ($stackPtr + 1));
         }
+
         $expression = '/(?i)(?=(@to(-|\s|)+do))(?-i)(?!@todo\s\w)/m';
         if ((bool) preg_match($expression, $comment) === true) {
             $phpcsFile->addError('@todo comments should be in the "@todo Some task." format.', $stackPtr, 'TodoFormat');
--- a/coder_sniffer/Drupal/Sniffs/Commenting/TodoCommentSniff.php
+++ b/coder_sniffer/Drupal/Sniffs/Commenting/TodoCommentSniff.php
@@ -53,12 +53,10 @@ class TodoCommentSniff implements Sniff
         $tokens = $phpcsFile->getTokens();
         // Standard comments and multi-line comments where the "@" is missing so
         // it does not register as a T_DOC_COMMENT_TAG.
-        if (in_array($tokens[$stackPtr]['code'], [T_COMMENT, T_DOC_COMMENT_STRING])) {
+        if ($tokens[$stackPtr]['code'] === T_COMMENT || $tokens[$stackPtr]['code'] === T_DOC_COMMENT_STRING) {
             $comment = $tokens[$stackPtr]['content'];
             $this->checkTodoFormat($phpcsFile, $stackPtr, $comment);
-        }
-
-        else if ($tokens[$stackPtr]['code'] === T_DOC_COMMENT_TAG) {
+        } else if ($tokens[$stackPtr]['code'] === T_DOC_COMMENT_TAG) {
             // Document comment tag (i.e. comments that begin with "@").
             // Determine if this is related at all and build the full comment line
             // from the various segments that the line is parsed into.
@@ -72,22 +70,22 @@ class TodoCommentSniff implements Sniff
                 }
 
                 $this->checkTodoFormat($phpcsFile, $stackPtr, $comment);
-            }
+            }//end if
         }
 
     }//end process()
 
 
-  /**
-   * Checks a comment string for the correct syntax.
-   *
-   * @param \PHP_CodeSniffer\Files\File $phpcsFile The file being scanned.
-   * @param int                         $stackPtr The position of the current token
-   *                                              in the stack passed in $tokens.
-   * @param string                      $comment The comment text.
-   *
-   * @return void
-   */
+    /**
+     * Checks a comment string for the correct syntax.
+     *
+     * @param \PHP_CodeSniffer\Files\File $phpcsFile The file being scanned.
+     * @param int                         $stackPtr  The position of the current token
+     *                                               in the stack passed in $tokens.
+     * @param string                      $comment   The comment text.
+     *
+     * @return void
+     */
     private function checkTodoFormat(File $phpcsFile, $stackPtr, string $comment)
     {
         $expression = '/^(\/\/|\*)*\s*(?i)(?=(@*to(-|\s|)+do))(?-i)(?!@todo\s(?!-|:))/m';
--- a/coder_sniffer/Drupal/Sniffs/Commenting/TodoCommentSniff.php
+++ b/coder_sniffer/Drupal/Sniffs/Commenting/TodoCommentSniff.php
@@ -57,7 +57,7 @@ class TodoCommentSniff implements Sniff
             $comment = $phpcsFile->findNext(T_DOC_COMMENT_STRING, ($stackPtr + 1));
         }
 
-        $expression = '/^\s*(?i)(?=(@*to(-|\s|)+do))(?-i)(?!@todo\s{1}[A-Z])/m';
+        $expression = '/^(\/\/)*\s*(?i)(?=(@*to(-|\s|)+do))(?-i)(?!@todo\s{1}[A-Z])/m';
         if ((bool) preg_match($expression, $comment) === true) {
             $phpcsFile->addError('@todo comments should be in the "@todo Some task." format.', $stackPtr, 'TodoFormat');
         }
--- a/coder_sniffer/Drupal/Sniffs/Commenting/TodoCommentSniff.php
+++ b/coder_sniffer/Drupal/Sniffs/Commenting/TodoCommentSniff.php
@@ -49,7 +49,7 @@ class TodoCommentSniff implements Sniff
      */
     public function process(File $phpcsFile, $stackPtr)
     {
-        $tokens  = $phpcsFile->getTokens();
+        $tokens = $phpcsFile->getTokens();
         // Use the comment text by default.
         $comment = $tokens[$stackPtr]['content'];
         // Use the next line for multi-line comments.
--- a/coder_sniffer/Drupal/Sniffs/Commenting/TodoCommentSniff.php
+++ b/coder_sniffer/Drupal/Sniffs/Commenting/TodoCommentSniff.php
@@ -90,7 +90,8 @@ class TodoCommentSniff implements Sniff
     {
         $expression = '/^(\/\/|\*)*\s*(?i)(?=(@*to(-|\s|)+do))(?-i)(?!@todo\s(?!-|:))/m';
         if ((bool) preg_match($expression, $comment) === true) {
-            $phpcsFile->addError('@todo comments should be in the "@todo Some task." format.', $stackPtr, 'TodoFormat');
+            $comment = trim($comment, " /\r\n");
+            $phpcsFile->addError("'%s' should match the format '@todo Some task'", $stackPtr, 'TodoFormat', [$comment]);
         }
 
     }//end checkTodoFormat()
--- a/coder_sniffer/Drupal/Sniffs/Commenting/TodoCommentSniff.php
+++ b/coder_sniffer/Drupal/Sniffs/Commenting/TodoCommentSniff.php
@@ -1,0 +1,65 @@
+<?php
+/**
+ * Parses and verifies comment language.
+ *
+ * @category PHP
+ * @package  PHP_CodeSniffer
+ * @link     http://pear.php.net/package/PHP_CodeSniffer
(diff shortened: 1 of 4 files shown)


jonathan1055:
Nice work. Will review shortly. I know that Klausi will want test coverage for this too.

coder_sniffer/Drupal/Sniffs/Commenting/TodoCommentSniff.php
```diff
@@ -0,0 +1,68 @@
...
+     *
+     * @return void
+     */
+    public function process(File $phpcsFile, $stackPtr)
+    {
+        $tokens = $phpcsFile->getTokens();
+        // Use the comment text by default.
+        $comment = $tokens[$stackPtr]['content'];
+        // Use the next line for multi-line comments.
+        if ($tokens[$stackPtr]['code'] === T_DOC_COMMENT_TAG) {
```
jonathan1055: I used `$tokens[$stackPtr]['type']` here. I don't know which is preferred.

coder_sniffer/Drupal/Sniffs/Commenting/TodoCommentSniff.php
```diff
@@ -0,0 +1,68 @@
...
+            $comment = $phpcsFile->findNext(T_DOC_COMMENT_STRING, ($stackPtr + 1));
+        }
+
+        $expression = '/^\s*(?i)(?=(@*to(-|\s|)+do))(?-i)(?!@todo\s{1}[A-Z])/m';
+        if ((bool) preg_match($expression, $comment) === true) {
+            $phpcsFile->addError('@todo comments should be in the "@todo Some task." format.', $stackPtr, 'TodoFormat');
```
jonathan1055: You could have
````
$comment = trim($comment, " /\r\n");
$phpcsFile->addError("'%s' should match the format '@todo Some task'", $stackPtr, 'TodoFormat', [$comment]);
````
Then the user sees what is wrong. This also requires that the tag text is appended to the original $comment, so line 57 would become
````
$pointer  = $phpcsFile->findNext(T_DOC_COMMENT_STRING, ($stackPtr + 1));
$comment .= ' '.$tokens[$pointer]['content'];
````

**Comment by jonathan1055** [comment]
Did you actually run your latest changes? I get nothing reported at all, for both types of bad format. I use
````
phpcs --standard=Drupal -vs --colors --sniffs=Drupal.Commenting.TodoComment path/to/testfile.php
````

**Comment by jonathan1055** [comment]
>     * Will the `// ` be in the string that is sent to the sniff? If so, I need to adjust it.
Yes `//` will be in the content for T_COMMENT, but in the not in the string returned by `findNext(T_DOC_COMMENT_STRING`

>     * Should we enforce a capital letter after `@todo`? (I went ahead and did this.)
I'd say yes, and you also enforced one space. These are all good, but I don't know if they are actually written in the standard?

File: coder_sniffer/Drupal/Sniffs/Commenting/TodoCommentSniff.php (hunk 1 above)
Comment (jonathan1055) [reply coder_sniffer/Drupal/Sniffs/Commenting/TodoCommentSniff.php]: `$tokens[$stackPtr]['code'] === T_DOC_COMMENT_TAG` will never be true. It should be either 
````
$tokens[$stackPtr]['type'] === T_DOC_COMMENT_TAG
````
or
````
$tokens[$stackPtr]['code'] === 'PHPCS_T_DOC_COMMENT_TAG'
````
The 'type' check I think is preferrable, because they are always readable. Sometimes 'code' is a number, other times it is a string as above, whereas checking 'type' can always use the constants and will be consistent.

File: coder_sniffer/Drupal/Sniffs/Commenting/TodoCommentSniff.php [hunk 3]
```diff
@@ -0,0 +1,68 @@
...
+     * @return void
+     */
+    public function process(File $phpcsFile, $stackPtr)
+    {
+        $tokens = $phpcsFile->getTokens();
+        // Use the comment text by default.
+        $comment = $tokens[$stackPtr]['content'];
+        // Use the next line for multi-line comments.
+        if ($tokens[$stackPtr]['code'] === T_DOC_COMMENT_TAG) {
+            $comment = $phpcsFile->findNext(T_DOC_COMMENT_STRING, ($stackPtr + 1));
```
jonathan1055: The above will give the pointer not the comment text. You need something like
````
$pointer = $phpcsFile->findNext(T_DOC_COMMENT_STRING, ($stackPtr + 1));
$comment = $tokens[$pointer]['content'];
````

**Comment by adamzimmermann** [comment]
@jonathan1055 thanks for all of the great feedback. 🙏 

> Did you actually run your latest changes?

I ran them a few regex iterations ago. I wanted to get a review on the high level approach first. Clearly I need to do some actual testing again though.

I'll work on some fixes in the next couple days and report back.

File: coder_sniffer/Drupal/Sniffs/Commenting/TodoCommentSniff.php (hunk 2 above)
Comment (jonathan1055) [reply coder_sniffer/Drupal/Sniffs/Commenting/TodoCommentSniff.php]: Actually, we need to do the appending above regardless of whether we use the $comment substitution in the message. It is needed because the `@whatever` tag will be in the original $comment variable, and this is needed in the final text that is used in the preg_match. May need to look at exactly how the appending is done, there could be multiple spaces in the source file, which might get lost, or may need to trim `$tokens[$pointer]['content']` so that we don't end up with more spaces that the original source. But these details can wait until the rest of it is working.

File: coder_sniffer/Drupal/Sniffs/Commenting/TodoCommentSniff.php (hunk 1 above)
Comment (adamzimmermann) [reply coder_sniffer/Drupal/Sniffs/Commenting/TodoCommentSniff.php]: Other sniffs seem to use the `T_DOC_*` constants, which I assume is preferable to the `'PHPCS_T_DOC_COMMENT_TAG'` string? Especially since that string is simple the value of that constant.

```
// Tokens used for parsing doc blocks.
define('T_DOC_COMMENT_STAR', 'PHPCS_T_DOC_COMMENT_STAR');
define('T_DOC_COMMENT_WHITESPACE', 'PHPCS_T_DOC_COMMENT_WHITESPACE');
define('T_DOC_COMMENT_TAG', 'PHPCS_T_DOC_COMMENT_TAG');
define('T_DOC_COMMENT_OPEN_TAG', 'PHPCS_T_DOC_COMMENT_OPEN_TAG');
define('T_DOC_COMMENT_CLOSE_TAG', 'PHPCS_T_DOC_COMMENT_CLOSE_TAG');
define('T_DOC_COMMENT_STRING', 'PHPCS_T_DOC_COMMENT_STRING');
```

Thoughts?

coder_sniffer/Drupal/Sniffs/Commenting/TodoCommentSniff.php
adamzimmermann: Figuring this all out took some thinking, as I'm new to writing code sniffs, but I think I sorted it out. Thanks for the guidance here.

jonathan1055:
Nice work. I started looking at your changes, but now there are lots, I think I will wait until you say you are ready for review. Let me know.

adamzimmermann:
I was hoping this last commit would fix all of the remaining Travis issue, but it failed to build. I don't see a re-run tests option. So I might need some assistance here.

Either way, I believe this is ready for review.

The only other thing I can think to add is the ability to fix the issue automatically, but that might be a follow up PR.

jonathan1055:
> I was hoping this last commit would fix all of the remaining Travis issue, but it failed to build. I don't see a re-run tests option. So I might need some assistance here.

I created a local branch with all your changes and tried  ` vendor/phpunit/phpunit/phpunit -v --filter=todo`
Yes there was a big problem, nothing happened! apart from the fan on my laptop starting to whirr. Ctnl C.

So I looked at the test file, and noticed in TodoCommentUnitTest.php `getErrorList()` your definition of $errorList had
`for ($i = 25; $i < 100; $i + 4) {` and the problem was this loop never finished. You need `$i += 4`
Then the tests at least ran but gave many failures. 

Your recent change to split all the incorrect docbock @todo into a separate comment block (not sure why you did that) but anyway, these are not being found as valid docblocks because they only have `/*` as the first row, they all need `/**`. With these two changes most of the tests passed. I suggest we make just these two fixes, and push a commit here so we can see the test run. Then we can work on the remainder of the review and the test failures.  I may be able to push to this PR, I don't know.

> The only other thing I can think to add is the ability to fix the issue automatically, but that might be a follow up PR.

Definitely leave that until the main work is done.

coder_sniffer/Drupal/Sniffs/Commenting/TodoCommentSniff.php
jonathan1055: Yes the `tokens[$stackPtr]['type'] === T_DOC_*` way is preferrable, and this is what you've done.

adamzimmermann:
> I created a local branch with all your changes and tried vendor/phpunit/phpunit/phpunit -v --filter=todo Yes there was a big problem, nothing happened! apart from the fan on my laptop starting to whirr. Ctnl C.

That's not what I wanted haha. I was running the sniffs locally, but wasn't running the tests locally and it shows haha. I'll get the tests running locally too as your next comment indicates there are still issues beyond the big one.

> So I looked at the test file, and noticed in TodoCommentUnitTest.php getErrorList() your definition of $errorList had
for ($i = 25; $i < 100; $i + 4) { and the problem was this loop never finished. You need $i += 4
Then the tests at least ran but gave many failures.

🤦 That would do it.

> Your recent change to split all the incorrect docbock @todo into a separate comment block (not sure why you did that) but anyway, these are not being found as valid docblocks because they only have /* as the first row, they all need /**. With these two changes most of the tests passed. I suggest we make just these two fixes, and push a commit here so we can see the test run. Then we can work on the remainder of the review and the test failures. I may be able to push to this PR, I don't know.

I had to do that to make the code alignment sniff happy. Some of the malformed statements were causing it to bark about how other things were indented, so I separated them. I was struggling with the formatting as the standards enforced in this repo are very different from the code style I'm used to working with.

I'll try making the comments use `/**`, but I seem to remember it barking about some other formatting issues then.

I was seeing `Inline doc block comments are not allowed; use "/* Comment */" or "// Comment" instead`. I added a bunch of functions to get around this.

> Definitely leave that until the main work is done.

👍 

I'll get some updates pushed in the next couple days (maybe later today 🤞).

adamzimmermann:
Everything seems to be passing. ✅

jonathan1055:
>I had to do that to make the code alignment sniff happy. Some of the malformed statements were causing it to bark about how other things were indented, so I separated them. I was struggling with the formatting as the standards enforced in this repo are very different from the code style I'm used to working with.

Yes I saw in your commits you were having to work at it. However, the test files are not actually part of the standard phpcs scan so they are not subject to this checking. But unfortunately you had named yours `TodoCommentUnitTest.inc.php` so it was being included, and hence caused you trouble. The test files should be named .inc without the .php, then coder's phpcs.xml.dist file with ignore them due to
````
    <!-- Test files that should not be checked. -->
    <exclude-pattern>tests/*\.(inc|css|js|api\.php|tpl\.php)$</exclude-pattern>
````

So, you can return to a simple test files, and I've done this in my copy of your branch, to save you the trouble - see
https://github.com/jonathan1055/coder/tree/118-todo-17
You should be able to merge those changes into yours using https://github.com/adamzimmermann/coder/compare/8.x-3.x...jonathan1055:118-todo-17?expand=1 or I can provide a patch to take you from your current state (after 19 commits) up to this latest version if you want it. I slightly modified a couple of the valid test cases, to show that $ and \ are accepted as first char of the todo text.  I also added the standard method of producing debug, so we can run on a small test file and get detailed output locally of what is happening.

I have a couple more questions but lets get the simple test file merged into your branch first.

jonathan1055:
I have made a few more suggestions and pushed them to my branch. 

1. I have reformatted the regex to allow freespace and commenting. This is much easier to understand and alter
1. T_COMMENT matches ordinary comments in // and this includes the // so we do need to cater for // in the regex. Howvwre I have allowed for any number in case we get /// etc
1. T_DOC_COMMENT_STRING find all comments lines in a docblock. This _does not_ inlcude the starting * so there no need to cater for that in the regex
1. T_DOC_COMMENT_TAG finds all doc block tags, i.e. @anything within a /** doc block. You correctly build up the the rest of the line of text in a loop, to ensure we get the spacing accurate. Unfortunately this text will also be found later in the procesing because it matches T_DOC_COMMENT_STRING but it will not be reported twice as the @todo wont be found. It's just a minor ineficiency that it gets tested twice.
1. Just for the record @to-do, @to - do, @to- ---  do are all considered to be @todo tags and are therefore checked (and failed)
1. @to and @to- are not considered to be @todo tags, and are not checked, which is correct.
1. If there is no @ in front of the todo then I think we should only count todo and to-do as valid @todo tags to be tested. Otherwise we would match and report on the second line of the following comment which is clearly not a todo tag:
````
````
I have made this change in the regex, and included test coverage.
8. Given that this @todo format is only advisory, should these all be coding standards warnings not errors?

Overall excelent work, this will be a good addition. I hope you can easily compare my branch to yours and get the changes? If you want patches let me know.

adamzimmermann:
Hoping to find time to get back to this today. Thanks for the contributions!

jonathan1055:
OK. If you want to discuss anything, I have just started using Slack. The coder channel is https://app.slack.com/client/T06GX3JTS/CJ19MV96J

adamzimmermann:
👍 What is the subdomain?

Screen Shot 2020-09-29 at 8 46 30 AM

jonathan1055:
drupal.slack.com 
I can see your id, and just added you to #coder channel

adamzimmermann:
Closing in favor of https://github.com/pfrenssen/coder/pull/120.

jonathan1055:
You made the comment:
_"I was struggling with the formatting as the standards enforced in this repo are very different from the code style I'm used to working with."_

I forgot to say at the time that, yes, this standard is not what I'm used to. However, running `phpcbf` can automatically fix a large proportion of the formatting requirements that we often miss, and really eases the hassle out of it.

PR #119, opened 2020-09-18
Use $parenthesisCloser column not the full line length

jonathan1055:
Durpal.org issue https://www.drupal.org/project/coder/issues/3169452

When calculating the lie length of inline array, use `$tokens[$tokens[$stackPtr][$parenthesisCloser]]['column']` not the full line length. Apologies to @klausi he did question this in my fix on [PR 114](https://github.com/pfrenssen/coder/pull/114)

This first commit fixes the code, but does not change the tests, and there will be one failure. This is where the array ends within the limit but the line extends beyond. I had marked this as a fault, but now it will not be.


arkener:
Makes sense, thanks!

jonathan1055:
Thanks.

PR #120, opened 2020-10-07
Create @todo format sniff.

adamzimmermann:
Adds @todo format sniff.

Drupal.org issue:
- https://www.drupal.org/project/coder/issues/2222435

Testing/Examples:
- https://regex101.com/r/cir1XP/9

Questions:
~~Will the `// ` be in the string that is sent to the sniff? If so, I need to adjust it.~~
~~- Should we enforce a capital letter after `@todo`? (I went ahead and did this.)~~

---

Closing #118 in favor of this PR to simplify/avoid merge conflicts from [this PR](https://github.com/adamzimmermann/coder/pull/1) where the following was done:

> I have made a few more suggestions and pushed them to my branch.

> I have reformatted the regex to allow freespace and commenting. This is much easier to understand and alter
T_COMMENT matches ordinary comments in // and this includes the // so we do need to cater for // in the regex. Howvwre I have allowed for any number in case we get /// etc
T_DOC_COMMENT_STRING find all comments lines in a docblock. This does not inlcude the starting * so there no need to cater for that in the regex
T_DOC_COMMENT_TAG finds all doc block tags, i.e. @anything within a /** doc block. You correctly build up the the rest of the line of text in a loop, to ensure we get the spacing accurate. Unfortunately this text will also be found later in the procesing because it matches T_DOC_COMMENT_STRING but it will not be reported twice as the @todo wont be found. It's just a minor ineficiency that it gets tested twice.
Just for the record @to-do, @to - do, @to- --- do are all considered to be @todo tags and are therefore checked (and failed)
@to and @to- are not considered to be @todo tags, and are not checked, which is correct.
If there is no @ in front of the todo then I think we should only count todo and to-do as valid @todo tags to be tested. Otherwise we would match and report on the second line of the following comment which is clearly not a todo tag:
I have made this change in the regex, and included test coverage.
> 8. Given that this @todo format is only advisory, should these all be coding standards warnings not errors?

> Overall excelent work, this will be a good addition. I hope you can easily compare my branch to yours and get the changes? If you want patches let me know.


tests/Drupal/Commenting/TodoCommentUnitTest.php
```diff
@@ -38,7 +37,8 @@ protected function getErrorList(string $testFile): array
      */
     protected function getWarningList(string $testFile): array
     {
-        return [];
+        $errorList = (array_fill_keys(range(13, 31), 1) + array_fill_keys(range(45, 63), 1));
```
jonathan1055: You could change the variable name to $warningList so it is clearer

jonathan1055:
Tests pass, that's great.

tests/Drupal/Commenting/TodoCommentUnitTest.php
adamzimmermann: Good call. Changing.

adamzimmermann:
I believe this is ready for final review and merging. @jonathan1055 thanks for all of your feedback and help on this.

jonathan1055:
Yes this all looks good. I think you need to add a new comment on the drupal issue (a) to link to this new PR and (b) to highlight that it is ready for @Arkener or @klausi to review..

coder_sniffer/Drupal/Sniffs/Commenting/TodoCommentSniff.php
```diff
@@ -0,0 +1,157 @@
...
+     * @param int                         $stackPtr  The position of the current token
+     *                                               in the stack passed in $tokens.
+     *
+     * @return void
+     */
+    public function process(File $phpcsFile, $stackPtr)
+    {
+        $debug = Config::getConfigData('todo_debug');
+        if ($debug !== null) {
+            $this->debug = (bool) $debug;
```
klausi, line 69: I think we should not initialize this variable on every single process() call? Can we do that once in the constructor instead?

Ah, I see that this approach is copied from ScopeIndentSniff, so also not ideal there. Anyway, then I think it is fine to keep this as is.

coder_sniffer/Drupal/Sniffs/Commenting/TodoCommentSniff.php
```diff
@@ -0,0 +1,157 @@
...
+              (?!-|:)\S        # and then any non-space except - or :
+            )/m';
+
+        if ((bool) preg_match($expression, $comment) === true) {
+            if ($this->debug === true) {
+                echo "Failed regex - give message\n";
+            }
+
+            $comment = trim($comment, " /\r\n");
+            $phpcsFile->addWarning("'%s' should match the format '@todo Some task'", $stackPtr, 'TodoFormat', [$comment]);
```
klausi, line 151: Comments should end with a dot, also todo comments. So the example should be something like `@todo Fix problem X here.`

klausi: It would be really cool to have a fixer here, that would help many people auto-fix this standard (also in Drupal core). I think it should be possible, we should have all info in the regex and could try a preg_replace().

Not a blocker for this pull request, we can merge this as first step.

klausi (commented):
Nice, thanks a lot!

I will do a test run against Drupal core to check if I see any false positives.

coder_sniffer/Drupal/Sniffs/Commenting/TodoCommentSniff.php
jonathan1055: Yes and I also used it in Commenting/DeprecatedSniff Would you like me to start a new issus to change all three of these, when this is committed?

klausi:
All warnings found in Drupal core look correct to me, yay! Will merge this now and improve the warning message as I proposed.

coder_sniffer/Drupal/Sniffs/Commenting/TodoCommentSniff.php
jonathan1055: Yes, doc comments not ending with a period is checked for. But currently @ tags are not considered within that group, so no warning is given if they have no period at the end. I agree that if they should end in a dot then this example should include the dot. But then we should also include @ tags in the general sniff for no ending period. @see should not be included, we discussed that on the deprecation sniffs, the url should not end in a dot. Maybe this is for another issue?

Regarding a fixer, yes it should be possible. @ adamzimmermann did suggest that, but we wanted to wait until we knew you were interested in this sniff

klausi: I see, very considerate of you! Yep, I absolutely support this sniff.

Also yep to the ending dot discussion, should be a separate issue.

jonathan1055: > Also yep to the ending dot discussion, should be a separate issue.

I have raised https://www.drupal.org/project/coder/issues/3183156 to check for doc tags not ending with a period.

PR #122, opened 2020-10-19
fix(InlineVariableComment): Ignore commented @var doc comments to fix nested doc comments

arkener:
Reference: https://www.drupal.org/project/coder/issues/3177757

### DIFF
--- a/coder_sniffer/Drupal/Sniffs/Commenting/InlineVariableCommentSniff.php
+++ b/coder_sniffer/Drupal/Sniffs/Commenting/InlineVariableCommentSniff.php
@@ -83,16 +83,18 @@ class InlineVariableCommentSniff implements Sniff
             if (strpos($tokens[$stackPtr]['content'], '@var') !== false) {
                 $warning = 'Inline @var declarations should use the /** */ delimiters';
 
-                if ($phpcsFile->addFixableWarning($warning, $stackPtr, 'VarInline') === true) {
-                    if (strpos($tokens[$stackPtr]['content'], '#') === 0) {
-                        $varComment = substr_replace(rtrim($tokens[$stackPtr]['content']), '/**', 0, 1)." */\n";
-                    } else if (strpos($tokens[$stackPtr]['content'], '//') === 0) {
-                        $varComment = substr_replace(rtrim($tokens[$stackPtr]['content']), '/**', 0, 2)." */\n";
-                    } else {
-                        $varComment = substr_replace($tokens[$stackPtr]['content'], '/**', 0, 2);
+                if (strpos($tokens[$stackPtr]['content'], '#') === 0 || strpos($tokens[$stackPtr]['content'], '//') === 0) {
+                    if (strpos($tokens[$stackPtr]['content'], '*/') !== false) {
+                        return;
                     }
 
-                    $phpcsFile->fixer->replaceToken($stackPtr, $varComment);
+                    if ($phpcsFile->addFixableWarning($warning, $stackPtr, 'VarInline') === true) {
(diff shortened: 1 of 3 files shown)


arkener:
Unsure why the tests results aren't shown here, but they're available on https://travis-ci.org/github/pfrenssen/coder/builds/737090959

tests/Drupal/Commenting/InlineVariableCommentUnitTest.inc
```diff
@@ -16,5 +16,9 @@ function lorem() {
   // @var $node \Drupal\node\NodeInterface
   # @var \Drupal\node\NodeInterface $node
   # @var $node \Drupal\node\NodeInterface
+  // /** @var \Drupal\node\NodeInterface $node */
+  // /* @var \Drupal\node\NodeInterface $node */
+  # /* @var \Drupal\node\NodeInterface $node */
+  # /* @var \Drupal\node\NodeInterface $node */
```
jonathan1055, line 22: Is there a reason for this repeated line? Maybe there is a something subtle that I am missing?

coder_sniffer/Drupal/Sniffs/Commenting/InlineVariableCommentSniff.php
```diff
@@ -83,16 +83,18 @@ public function process(File $phpcsFile, $stackPtr)
...
                 $warning = 'Inline @var declarations should use the /** */ delimiters';
 
-                if ($phpcsFile->addFixableWarning($warning, $stackPtr, 'VarInline') === true) {
-                    if (strpos($tokens[$stackPtr]['content'], '#') === 0) {
-                        $varComment = substr_replace(rtrim($tokens[$stackPtr]['content']), '/**', 0, 1)." */\n";
-                    } else if (strpos($tokens[$stackPtr]['content'], '//') === 0) {
-                        $varComment = substr_replace(rtrim($tokens[$stackPtr]['content']), '/**', 0, 2)." */\n";
-                    } else {
-                        $varComment = substr_replace($tokens[$stackPtr]['content'], '/**', 0, 2);
+                if (strpos($tokens[$stackPtr]['content'], '#') === 0 || strpos($tokens[$stackPtr]['content'], '//') === 0) {
```
jonathan1055, line 86: I find this harder to read than the original. Maybe add a short comment? Also creating variable `$varComment` was clear before, but now you just do the replaceToken() directly with nested rtrim/ltrim. So a source comment explaining it would be very helpful

tests/Drupal/Commenting/InlineVariableCommentUnitTest.inc
arkener: Good catch, forgot to add an extra `*` to the first line.

coder_sniffer/Drupal/Sniffs/Commenting/InlineVariableCommentSniff.php
arkener: Fair point, I've re-added the `$varComment` variables the to replace the one-liner and added comments why the trims are needed.

arkener:
Thanks for the review @jonathan1055, I've updated the PR with more comments and fixed the duplicate test line.

coder_sniffer/Drupal/Sniffs/Commenting/InlineVariableCommentSniff.php
```diff
@@ -83,18 +83,27 @@ public function process(File $phpcsFile, $stackPtr)
...
+                    if (strpos($tokens[$stackPtr]['content'], '*/') !== false) {
```
klausi, line 87: Did not understand this at first, but we should leave a code comment above like "// If this comment contains '*/' then the developer is mixing inline comment styles. This could be commented out code, so leave this line alone completely."

klausi (approved):
Nice! I was a bit confused and would like to see a code comment, but then this can be merged!

PR #123, opened 2020-10-20
Add @todo Fixing Functionality

adamzimmermann:
https://www.drupal.org/project/coder/issues/3177471


adamzimmermann:
All tests and code style checks are passing.

adamzimmermann:
I'm not sure why this is showing old commits. If you are squashing when you merge, it shouldn't matter though I suppose. If you want me to open a new PR that only shows the new commits I can do that too. The diff looks correct though.

arkener (changes requested):
Thank you for working on this! The PHPStan [check](https://travis-ci.org/github/pfrenssen/coder/jobs/737534829) currently seems to be failing, I'm unsure why the check isn't showing up here, @klausi any idea? 

Running this on core returns the expected results with 1 exception. The fixer currently breaks when an empty `@todo` is given, for example:
```php
function lorem() {
  // @todo
  $node = Node::load(1);
}
```
Results in the following
```php
function lorem() {
  // @todo $node = Node::load(1);
}
```

Could you add a test for the fixer? That way we can automatically validate the fixer. This can be achieved by adding the `TodoCommentUnitTest.inc.fixed` file, containing the expected result of the autofix based on the `TodoCommentUnitTest.inc` file.

klausi:
Strange that Travis CI does not show up. I also see some other commits here - can you make a new pull request where you start the branch fresh from origin/8.x-3.x?

adamzimmermann:
New PR with those fixes coming up.

adamzimmermann:
I fixed the new line creation, but how do we want to handle `// @todo`? Should that remain valid? i.e. we don't want to flag it at all? If we do flag it, we can't fix it unless we add some example text.

Maybe it should be fixed to `@todo Issue description here.`?

PR #127, opened 2020-10-29
Ensure module dependencies are an array. D.O #issue_id 3166470

branjansdl1:
d.o. issue https://www.drupal.org/node/3166470

### DIFF
--- a/coder_sniffer/Drupal/Sniffs/InfoFiles/DependenciesArraySniff.php
+++ b/coder_sniffer/Drupal/Sniffs/InfoFiles/DependenciesArraySniff.php
@@ -10,15 +10,16 @@ namespace Drupal\Sniffs\InfoFiles;
 use PHP_CodeSniffer\Files\File;
 use PHP_CodeSniffer\Sniffs\Sniff;
 use Symfony\Component\Yaml\Yaml;
+use Symfony\Component\Yaml\Exception\ParseException;
 
 /**
- * Dependency should have single entry as an array in Drupal 8 info files.
+ * Dependencies should have single entry as an array in Drupal 8 info files.
  *
  * @category PHP
  * @package  PHP_CodeSniffer
  * @link     http://pear.php.net/package/PHP_CodeSniffer
  */
-class DependencyArraySniff implements Sniff
+class DependenciesArraySniff implements Sniff
 {
 
 
@@ -50,13 +51,15 @@ class DependencyArraySniff implements Sniff
         if ($fileExtension !== 'info') {
             return ($phpcsFile->numTokens + 1);
         }
-
-        $info = Yaml::parse(file_get_contents($phpcsFile->getFilename()));
-        if (isset($info['dependencies']) === true) {
-            if (is_array($info['dependencies']) === false) {
-                $error = '"dependencies" should be an array in the info yaml file';
-                $phpcsFile->addError($error, $stackPtr, 'Dependencies');
-            }
+        try {
+            $info = Yaml::parse(file_get_contents($phpcsFile->getFilename()));
+        } catch (ParseException $e) {
+            // If the YAML is invalid we ignore this file.
+            return ($phpcsFile->numTokens + 1);
+        }
+        if (isset($info['dependencies']) === true && is_array($info['dependencies']) === false) {
+            $error = '"dependencies" should be an array in the info yaml file';
+            $phpcsFile->addError($error, $stackPtr, 'Dependencies');
         }
 
         return ($phpcsFile->numTokens + 1);
--- a/coder_sniffer/Drupal/Sniffs/InfoFiles/DependenciesArraySniff.php
+++ b/coder_sniffer/Drupal/Sniffs/InfoFiles/DependenciesArraySniff.php
@@ -26,7 +26,7 @@ class DependenciesArraySniff implements Sniff
     /**
      * Returns an array of tokens this test wants to listen for.
      *
-     * @return arrayint|string
+     * @return array<int|string>
      */
     public function register()
     {
--- a/coder_sniffer/Drupal/Sniffs/InfoFiles/DependenciesArraySniff.php
+++ b/coder_sniffer/Drupal/Sniffs/InfoFiles/DependenciesArraySniff.php
@@ -27,7 +27,6 @@ class DependenciesArraySniff implements Sniff
      * Returns an array of tokens this test wants to listen for.
      *
      * @return arrayint|string
-     *   Returns array or string.
      */
     public function register()
     {
@@ -42,7 +41,7 @@ class DependenciesArraySniff implements Sniff
      * @param \PHP_CodeSniffer\Files\File $phpcsFile The file being scanned.
      * @param int                         $stackPtr  The position of the current token in the stack passed in $tokens.
      *
-     * @return int Returns integer.
+     * @return int
      */
     public function process(File $phpcsFile, $stackPtr)
     {
@@ -51,12 +50,14 @@ class DependenciesArraySniff implements Sniff
         if ($fileExtension !== 'info') {
             return ($phpcsFile->numTokens + 1);
         }
+
         try {
             $info = Yaml::parse(file_get_contents($phpcsFile->getFilename()));
         } catch (ParseException $e) {
             // If the YAML is invalid we ignore this file.
             return ($phpcsFile->numTokens + 1);
         }
+
         if (isset($info['dependencies']) === true && is_array($info['dependencies']) === false) {
             $error = '"dependencies" should be an array in the info yaml file';
             $phpcsFile->addError($error, $stackPtr, 'Dependencies');
--- a/coder_sniffer/Drupal/Sniffs/InfoFiles/DependencyArraySniff.php
+++ b/coder_sniffer/Drupal/Sniffs/InfoFiles/DependencyArraySniff.php
@@ -1,0 +1,66 @@
+<?php
+/**
+ * \Drupal\Sniffs\InfoFiles\DependencyArraySniff.
+ *
+ * @category PHP
+ * @package  PHP_CodeSniffer
+ * @link     http://pear.php.net/package/PHP_CodeSniffer
+ */
+
+namespace Drupal\Sniffs\InfoFiles;
+
+use PHP_CodeSniffer\Files\File;
+use PHP_CodeSniffer\Sniffs\Sniff;
+use Symfony\Component\Yaml\Yaml;
+
+/**
+ * "dependency" should have single entry as an array in Drupal 8 info files.
+ *
+ * @category PHP
+ * @package  PHP_CodeSniffer
+ * @link     http://pear.php.net/package/PHP_CodeSniffer
+ */
+class DependencyArraySniff implements Sniff {
+
+
+  /**
+   * Returns an array of tokens this test wants to listen for.
+   *
+   * @return array<int|string>
+   */
+  public function register() {
+    return [T_INLINE_HTML];
+
+  }//end register()
+
+
+  /**
+   * Processes this test, when one of its tokens is encountered.
+   *
+   * @param \PHP_CodeSniffer\Files\File $phpcsFile The file being scanned.
+   * @param int $stackPtr The position of the current token in the
+   *                                               stack passed in $tokens.
+   *
+   * @return int
+   */
+  public function process(File $phpcsFile, $stackPtr) {
+    // Only run this sniff once per info file.
+    $fileExtension = strtolower(substr($phpcsFile->getFilename(), -8, 4));
+    if ($fileExtension !== 'info') {
+      return ($phpcsFile->numTokens + 1);
+    }
+
+    $info = Yaml::parse(file_get_contents($phpcsFile->getFilename()));
+    if (isset($info['dependencies']) === TRUE) {
+      if (!is_array($info['dependencies'])) {
+        $error = '"dependencies" should be an array in the info yaml file';
+        $phpcsFile->addError($error, $stackPtr, 'Dependencies');
+      }
+    }
+
+    return ($phpcsFile->numTokens + 1);
+
+  }//end process()
+
+
+}//end class
--- a/coder_sniffer/Drupal/Sniffs/InfoFiles/DependencyArraySniff.php
+++ b/coder_sniffer/Drupal/Sniffs/InfoFiles/DependencyArraySniff.php
@@ -1,5 +1,10 @@
 <?php
-
+/**
+ * Largely copied from
+ * PHP_CodeSniffer\Standards\Squiz\Sniffs\Classes\ClassFileNameSniff.
+ *
+ * Extended to support anonymous classes and Drupal core version.
+ */
 namespace Drupal\Sniffs\InfoFiles;
 
 use PHP_CodeSniffer\Files\File;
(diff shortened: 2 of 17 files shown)


jonathan1055:
Here's an active link to the d.o. issue https://www.drupal.org/node/3166470

jonathan1055:
Hi branjansdl1,
The Travis tests have all failed, but only due to coding standards messages in your new file. The standards for Coder module are a bit different from the Drupal standards, but you can easily run `phpcbf` from within the Coder root directory to automatically fix 27 of your 28 "faults"

branjansdl1:
I have run phpcs and found 16 cs issues for coder_sniffer/Drupal/Sniffs/InfoFiles/DependencyArraySniff.php file and i have fixed 13. Please review again i have pushed the code.

jonathan1055:
Hi,
I'm not sure which standards you are using when fixing, but unfortunately you are going the wrong way, there are now 38 standards messages, when before there were only 28. If you use `phpcs --standard=phpcs.xml.dist` this will ensure you use the standards as defined for Coder. Then `phpcbf --standard=phpcs.xml.dist` to fix them automatically.

jonathan1055:
I suggest you go back to your original commit, which was clearly based on a copy of an existing sniff, and has a lot of the correct formatting, which you have subsequently changed unnecessarily.

branjansdl1:
Thank you Jonathan for the guidance. I am using Drupal standards to fix these. Can you please suggest how to run coder specific standards inside coder root. It will be really helpful.

branjansdl1:
Apologies jonathan. i didnt check your previous comment. i will do the changes you have suggested and push the changes.

branjansdl1:
I have fixed the CS issues. Please review the updated code.

jonathan1055:
Yes, that is much better. Only 3 coding standards left to fix.
````
FILE: ...nssen/coder/coder_sniffer/Drupal/Sniffs/InfoFiles/DependencyArraySniff.php
--------------------------------------------------------------------------------
FOUND 3 ERRORS AFFECTING 3 LINES
--------------------------------------------------------------------------------
  2 | ERROR | Missing file doc comment
 36 | ERROR | Missing parameter comment
 38 | ERROR | Missing parameter comment
--------------------------------------------------------------------------------
````

You will also need to add test files for this new sniff, to show that the non-array dependency in the .yml file is correctly icentified, and that valid dependency is not flagged. You can follow example test files from other sniffs.

**Comment by jonathan1055** [comment]
The error on line 2 just requires that you add back what you had in the very first file
````
/**
 * \Drupal\Sniffs\InfoFiles\DependencyArraySniff.
 *
 * @category PHP
 * @package  PHP_CodeSniffer
 * @link     http://pear.php.net/package/PHP_CodeSniffer
 */
````

**Comment by jonathan1055** [comment]
and the errors on lines 36 and 38 are not very helpful. What it means is that the param comment is not found on the same line. Yours are on the next line. Look at `classFilesSniff.php` and you can see the param description on the same rows. You can just copy form that file.

**Comment by branjansdl1** [comment]
Thank you jonathan for your time on this. But i couldn't find any test case files to refer. There are 3 test files but in that also no asert or any logic found for functions. Could you please guide me on this?

**Comment by branjansdl1** [comment]
I have pushed the changes for the remaining cs fix. Please suggest how to add test case as i cant find any test case to refer.

**Comment by jonathan1055** [comment]
Yes the coder tests are different from the assert and php logic used in other contrib testing.

Look in the coder folder `tests/Drupal/InfoFiles` and you can see three files `AutoAddedKeysUnitTest.php`, `ClassFilesUnitTest.php` and `DuplicateEntryUnitTest.php`. These each check just one sniff, the sniff matching the name of the file and class. They use test files in the `drupal7` and `drupal8` subfolders there. 

I suggest you look at `AutoAddedKeysUnitTest.php` as that uses the drupal8/test.info.yml file (the other two tests only check the d7 .info file). Duplicate `AutoAddedKeysUnitTest.php` and call it `DependencyArrayUnitTest.php`. You will need to add an incorrect dependencies line in /drupal8/test.info.yml, then specify that the test should detect the error.

**Comment by jonathan1055** [comment]
If you want to make a start on the above, and see how you get on. You may be able to figure out the rest of it, as the comments in those files are quite good. 

To run the phpunit tests from the coder folder, use
```
phpunit -v --filter=InfoFiles
```
You may need to prefix that command with `sudo -u www` or something similar, depending on your local environment. Play around with the test file, and the array of expected errors and warnings. You will see how it is meant to work by intentionally making a change and watching how it fails. 

Ask any questions, I'm happy to help

**Comment by branjansdl1** [comment]
thank you so much for your help. really appreciated.

**Comment by branjansdl1** [comment]
Hi Jonathan, 
I have added tests for DependencyArraySniff can you please verify.

File: coder_sniffer/Drupal/Sniffs/InfoFiles/DependencyArraySniff.php [hunk 1]
```diff
@@ -0,0 +1,67 @@
...
+     * Processes this test, when one of its tokens is encountered.
+     *
+     * @param \PHP_CodeSniffer\Files\File $phpcsFile The file being scanned.
+     * @param int                         $stackPtr  The position of the current token in the stack passed in $tokens.
+     *
+     * @return int Returns integer.
+     */
+    public function process(File $phpcsFile, $stackPtr)
+    {
+        // Only run this sniff once per info file.
```
klausi: info file or info.yml file? It looks like you only go for Drupal 8+ yml files, right?

But I think this equally applies to Drupal 7 as well?

tests/Drupal/InfoFiles/drupal8/test.info.yml
```diff
@@ -5,3 +5,5 @@ name: Test
 project: test
 datestamp: 1234567
 version: 1.2
+dependency:
```
klausi: dependency or dependencies?

tests/Drupal/InfoFiles/drupal8/test.info.yml
```diff
@@ -5,3 +5,5 @@ name: Test
...
+  drupal.taxonomy
```
klausi: I think we should start a new info.yml test file in the drupal8 folder, then we don't mix test cases as much and they can be isolated better from each other. Please create a dependencies_array.info.yml test file.

tests/Drupal/InfoFiles/DependencyArraySniffUnitTest.php
```diff
@@ -0,0 +1,65 @@
+<?php
```
klausi: Wrong file name, should be DependencyArrayUnitTest.php

coder_sniffer/Drupal/Sniffs/InfoFiles/DependencyArraySniff.php
```diff
@@ -0,0 +1,67 @@
...
+use Symfony\Component\Yaml\Yaml;
+
+/**
+ * Dependency should have single entry as an array in Drupal 8 info files.
+ *
+ * @category PHP
+ * @package  PHP_CodeSniffer
+ * @link     http://pear.php.net/package/PHP_CodeSniffer
+ */
+class DependencyArraySniff implements Sniff
```
klausi: I think we should call this DependenciesArraySniff because the key in the info.yml file is "dependencies". Same for the test file.

coder_sniffer/Drupal/Sniffs/InfoFiles/DependencyArraySniff.php
```diff
@@ -0,0 +1,67 @@
...
+    }//end register()
+
+
+    /**
+     * Processes this test, when one of its tokens is encountered.
+     *
+     * @param \PHP_CodeSniffer\Files\File $phpcsFile The file being scanned.
+     * @param int                         $stackPtr  The position of the current token in the stack passed in $tokens.
+     *
+     * @return int Returns integer.
```
klausi: the comment just repeats what `@return int` says, so can be removed. You can leave that out as we do on other sniff classes.

coder_sniffer/Drupal/Sniffs/InfoFiles/DependencyArraySniff.php
```diff
@@ -0,0 +1,67 @@
...
+    {
+        // Only run this sniff once per info file.
+        $fileExtension = strtolower(substr($phpcsFile->getFilename(), -8, 4));
+        if ($fileExtension !== 'info') {
+            return ($phpcsFile->numTokens + 1);
+        }
+
+        $info = Yaml::parse(file_get_contents($phpcsFile->getFilename()));
+        if (isset($info['dependencies']) === true) {
+            if (is_array($info['dependencies']) === false) {
```
klausi: could be combined into one if() with &&

coder_sniffer/Drupal/Sniffs/InfoFiles/DependencyArraySniff.php
```diff
@@ -0,0 +1,67 @@
...
+     */
+    public function process(File $phpcsFile, $stackPtr)
+    {
+        // Only run this sniff once per info file.
+        $fileExtension = strtolower(substr($phpcsFile->getFilename(), -8, 4));
+        if ($fileExtension !== 'info') {
+            return ($phpcsFile->numTokens + 1);
+        }
+
+        $info = Yaml::parse(file_get_contents($phpcsFile->getFilename()));
```
klausi: we need to catch YAML parse exceptions that could occur, see AutoAddedKeysSniff

klausi (changes requested):
Thanks, looks pretty far already! I have a couple of comments that should be addressed.

coder_sniffer/Drupal/Sniffs/InfoFiles/DependencyArraySniff.php
branjansdl1: @klausi This is only for Drupal8+ as we have check for dependencies must have array in yml file.

tests/Drupal/InfoFiles/drupal8/test.info.yml
branjansdl1: Apologies. its dependencies.

tests/Drupal/InfoFiles/drupal8/test.info.yml
branjansdl1: I will create a new file for this.

tests/Drupal/InfoFiles/DependencyArraySniffUnitTest.php
branjansdl1: Thank you for correcting me. I will fix this.

coder_sniffer/Drupal/Sniffs/InfoFiles/DependencyArraySniff.php
branjansdl1: I will fix this.

coder_sniffer/Drupal/Sniffs/InfoFiles/DependencyArraySniff.php
branjansdl1: sure

tests/Drupal/InfoFiles/drupal8/test.info.yml
```diff
@@ -4,4 +4,4 @@ name: Test
 # These should not be here, will be added by drupal.org packaging.
 project: test
 datestamp: 1234567
-version: 1.2
+version: 1.2
```
klausi, line 7: This file should not be modified, please revert the change here.

tests/Drupal/InfoFiles/drupal8/dependencies_array.info.yml
```diff
@@ -0,0 +1,9 @@
+core: 8.x
+name: Dependencies Array Test
+
+# These should not be here, will be added by drupal.org packaging.
```
klausi, line 4: This and the next 3 lines can be removed for this test, we only want to test dependencies here, right?

klausi (changes requested):
Thanks, I have 2 minor comments, please fix those.

PHPCS coding standards are failing on Travis CI, please also fix those.

Otherwise I would say 99% finished here :)

tests/Drupal/InfoFiles/drupal8/test.info.yml
jonathan1055: @klausi I'd like to know why we cant use the exitsing `tests/Drupal/InfoFiles/drupal8/test.info.yml` and add the faults to it. The existing tests will all still pass on the changed file. I know sometimes that changing an existing test file causes problems in other tests, but it does not do so here. Or do you like having new test files for every sniff? The .info.yml tests are a bit different from the others. Just want to know our policy, for future enhancements. Thanks

jonathan1055 (commented):
asking advice on test file policy. Thanks

tests/Drupal/InfoFiles/drupal8/test.info.yml
jonathan1055: I did not see your comment above. OK maybe it is better to have separate test files.

klausi:
I like to separate test cases to isolate them better (I know we don't do that consistently across Coder, but we should try in the future). When we change sniffs and some tests fail it should be very clear why a related test fails.

jonathan1055:
Thanks, yes that is a good policy. After asking the question I read the full thead in detail and saw your previous explanation. Also, sorry I had not spotted the name problem (dependency when it should have been dependencies). I could have caught that one and commented for if to be fixed before you had to review.

klausi:
No worries, I also miss details when reviewing, that is normal I think. Thanks a lot for helping out reviewing pull requests!

jonathan1055:
That's the benefit of more than one person doing the review, we work as a team.

branjansdl1:
@klausi i have fixed the minor cs issues and pushed the code. Please have a check.

coder_sniffer/Drupal/Sniffs/InfoFiles/DependenciesArraySniff.php
```diff
@@ -0,0 +1,71 @@
...
+ * @link     http://pear.php.net/package/PHP_CodeSniffer
+ */
+class DependenciesArraySniff implements Sniff
+{
+
+
+    /**
+     * Returns an array of tokens this test wants to listen for.
+     *
+     * @return arrayint|string
```
klausi: data type should be `array<int|string>`

klausi:
Thanks, can you fix the open comments from me? Github has not auto-hidden those because the lines were not changed, so there are still a few things to do.

branjansdl1:
@klausi i have fixed the last commented issue. Please let me know if anything else pending on this.

tests/Drupal/InfoFiles/DependenciesArrayUnitTest.php
```diff
@@ -0,0 +1,65 @@
...
+     * The key of the array should represent the line number and the value
+     * should represent the number of warnings that should occur on that line.
+     *
+     * @param string $testFile The name of the file being tested.
+     *
+     * @return array<int, int>
+     */
+    protected function getWarningList(string $testFile): array
+    {
+        return [1 => 3];
```
klausi, line 44: tests are failing because this means "3 warnings on line 1". But your sniff does not produce warnings, it makes errors, right?

tests/Drupal/InfoFiles/DependenciesArrayUnitTest.php
```diff
@@ -0,0 +1,65 @@
...
+     * The key of the array should represent the line number and the value
+     * should represent the number of errors that should occur on that line.
+     *
+     * @param string $testFile The name of the file being tested.
+     *
+     * @return array<int, int>
+     */
+    protected function getErrorList(string $testFile): array
+    {
+        return [];
```
klausi, line 27: so you would need to specify "1 => 1" here (we don't support better line numbers in YAML files anyway)

tests/Drupal/InfoFiles/drupal8/dependencies_array.info.yml
klausi: can you address this?

tests/Drupal/InfoFiles/drupal8/test.info.yml
klausi: @branjansdl1 can you address this and not modify this file?

klausi (changes requested):
Thanks, tests are failing, left some inline comments.

klausi:
Closing this due to inactivity, feel free to open a new pull request if you want to continue!

jonathan1055:
I have created a new PR #186 to complete this work

PR #128, opened 2020-11-11
Doc tag group spacing

jonathan1055:
D.O. issue https://www.drupal.org/project/coder/issues/3181485

The first commit only changes the sniff. As discussed in the d.o. issue, the logic check for `$currentTag` and `$previousTag` being the the list of 'group' tags to check should be `||` not `&&` because it should not matter what the other tag is. This then meant that inline tags also became in scope for testing (which is wrong) so the column is now checked to make sure we are only dealing with real doc tags.

The test will fail, which highlight the fact that `@ingroup` is in the list to be checked but is only semi-fixed.

### DIFF
--- a/coder_sniffer/Drupal/Sniffs/Commenting/DocCommentSniff.php
+++ b/coder_sniffer/Drupal/Sniffs/Commenting/DocCommentSniff.php
@@ -373,6 +373,12 @@ class DocCommentSniff implements Sniff
         ];
         foreach ($tokens[$commentStart]['comment_tags'] as $pos => $tag) {
             if ($pos > 0) {
+                // If this tag is not in the same column as the initial tag then
+                // it must be an inline comment tag and should be ignored here.
+                if ($tokens[$tag]['column'] !== $tokens[$firstTag]['column']) {
+                    continue;
+                }
+
                 $prev = $phpcsFile->findPrevious(
                     T_DOC_COMMENT_STRING,
                     ($tag - 1),
@@ -387,7 +393,7 @@ class DocCommentSniff implements Sniff
                 if ($isNewGroup === true) {
                     $groupid++;
                 }
-            }
+            }//end if
 
             $currentTag = $tokens[$tag]['content'];
             if ($currentTag === '@param') {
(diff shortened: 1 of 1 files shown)


jonathan1055:
As expected we have the one test failure:
```
(1) Drupal\Test\Commenting\DocCommentUnitTest::testSniff
[LINE 17] Expected 0 error(s) in DocCommentUnitTest.inc but found 1 error(s). The error(s) found were:
 -> Separate the @ingroup and @deprecated sections by a blank line. (Drupal.Commenting.DocComment.TagGroupSpacing)

Fixed version of DocCommentUnitTest.inc does not match expected version in DocCommentUnitTest.inc.fixed; the diff is
--- tests/Drupal/../../tests/Drupal/Commenting/DocCommentUnitTest.inc.fixed
+++ PHP_CodeSniffer
@@ -18,6 +18,7 @@
  *   Thrown when $param is TRUE.
  *
  * @ingroup sniffer
+ *
  * @deprecated
  * @see some_function()
  * @todo Fix coding standards.
```
This shows that the sniff is working as intended, and because `@ingroup` is specified in our `$checkTags` array it is expecting that tag to be separated both before and after.

Now we can decide whether `@ingroup` should be in that list (in which case we add one blank line to the fixed file) or if it should not be separated, we remove it from the $checkTags array and the test 'fixed' file remains as is.

jonathan1055:
With reference to the earlier version of Coder and the current `DocCommentUnitTest.inc.fixed` file which has:
```
/**
 * Separate @param, @return, @throws sections by a blank line.
 *
 * @param string $param
 *   Something incredibly useful.
 *
 * @return bool
 *   Returns FALSE.
 *
 * @throws Exception
 *   Thrown when $param is TRUE.
 *
 * @ingroup sniffer
 * @deprecated
 * @see some_function()
 * @todo Fix coding standards.
 */
```
there is no evidence that `@ingroup` should be separated out with empty lines either side. It is just `@param`, `@return` and `@throws` that should be checked for blank lines either side. This is how it worked in Coder 8.3.6 and what was regressed in 8.3.7 so I have removed `@ingroup` from the array of tags to be checked for this.

With this correction there are now 12 cases in core 9.2 which fail the corrected sniff. These are probably lines that have been changed since the sniff was working, and they are all auto-fixable.

arkener:
Perfect!, thank you for working on this. The cases on 9.2.x which are currently correctly failing this sniff seem to be either related to incorrectly indented tags or actual incorrectly grouped tags. Merging this so these can be fixed in core.

PR #130, opened 2020-11-25
Improve fixer in ShortFullStop sniff

jonathan1055:
d.o. issue https://www.drupal.org/project/coder/issues/3184314
First change is to nly add full stop if short comment ends with alpah-numeric and is all contained on one line.

### DIFF
--- a/coder_sniffer/Drupal/Sniffs/Commenting/DocCommentSniff.php
+++ b/coder_sniffer/Drupal/Sniffs/Commenting/DocCommentSniff.php
@@ -233,6 +233,7 @@ class DocCommentSniff implements Sniff
         }
 
         $lastChar = substr($shortContent, -1);
+        // Allow these characters as valid line-ends not requiring to be fixed.
         if (in_array($lastChar, ['.', '!', '?', ')']) === false
             // Allow both variants of inheritdoc comments.
             && $shortContent !== '{@inheritdoc}'
@@ -242,9 +243,17 @@ class DocCommentSniff implements Sniff
             && $shortContent !== basename($phpcsFile->getFilename())
         ) {
             $error = 'Doc comment short description must end with a full stop';
-            $fix   = $phpcsFile->addFixableError($error, $shortEnd, 'ShortFullStop');
-            if ($fix === true) {
-                $phpcsFile->fixer->addContent($shortEnd, '.');
+            // If the last character is alphanumeric and the content is all on one line then fix it.
+            if (preg_match('/[a-zA-Z0-9]/', $lastChar) === 1
+                && $tokens[$short]['line'] === $tokens[$shortEnd]['line']
+            ) {
+                $fix = $phpcsFile->addFixableError($error, $shortEnd, 'ShortFullStop');
+                if ($fix === true) {
+                    $phpcsFile->fixer->addContent($shortEnd, '.');
+                }
+            } else {
+                // The correct fix is not obvious, so report an error and leave for manual correction.
+                $phpcsFile->addError($error, $shortEnd, 'ShortFullStop');
             }
         }
 
--- a/tests/Drupal/Commenting/DocCommentUnitTest.1.inc
+++ b/tests/Drupal/Commenting/DocCommentUnitTest.1.inc
@@ -7,3 +7,36 @@
  * A comprehensive guide with examples is available at:
  * @link http://example.com Test link @endlink.
  */
+
+/**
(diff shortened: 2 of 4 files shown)


jonathan1055:
I will expand the test coverage if this change is something that you'd consider to commit. It certainly was useful on core issues [3183673 ShortFullStop - auto-fixes](https://www.drupal.org/project/drupal/issues/3183673) and [3185652 ShortFullStop - manual follow-up](https://www.drupal.org/project/drupal/issues/3185652)

jonathan1055:
I think the reason this has 3 of Klausi's commits is because it is merging into `8.x-3.x`, so how can I change this to merge into `8.3.x`?

klausi (changes requested):
Thanks, I think that makes sense! Can you add all the test cases we want to improve that you mention in the drupal.org issue? Once we have the tests I think we can merge this.

jonathan1055:
Thanks @klausi glad you approve of the non-fixes. I did not see any extensive test coverage for this sniff, so I have added cases for 

1. where the comment is OK - ends in . ? ! or )
1. where it is not ok and should be fixed - ends in letter or digit
1. where it is not ok but cannot be reliably fixed - ending in wrong punctuation or spanning to multiple lines 

I used file `DocCommentUnitTest.1.inc` as this had minimal test cases in it, rather than add a new .4.inc file. I did not alter the existing test data, just added to the file.

tests/Drupal/Commenting/DocCommentUnitTest.1.inc
```diff
@@ -7,3 +7,36 @@
...
+/**
+ * Brackets are also an allowed ending with no warning()
+ */
+
+/**
+ * This is a short comment with no full stop that can be fixed automatically
+ */
+
+/**
+ * A comment ending in digits is fixable, but the number of these in Core is 0
```
klausi, line 28: lol, you are creative ... these tests make me smile :)

coder_sniffer/Drupal/Sniffs/Commenting/DocCommentSniff.php
```diff
@@ -242,9 +243,17 @@ public function process(File $phpcsFile, $stackPtr)
...
+            if (preg_match('/[a-zA-Z0-9]/', $lastChar) === 1
+                && $tokens[$short]['line'] === $tokens[$shortEnd]['line']
+            ) {
+                $fix = $phpcsFile->addFixableError($error, $shortEnd, 'ShortFullStop');
+                if ($fix === true) {
+                    $phpcsFile->fixer->addContent($shortEnd, '.');
+                }
+            } else {
+                // The correct fix is not obvious, so report an error and leave for manual correction.
+                $phpcsFile->addError($error, $shortEnd, 'ShortFullStop');
```
klausi, line 256: we could simplify this to have only one addFixableError() call and then check 3 conditions in the if(). Because we will trigger the error in any case, we just want to do different things if the fixer is active.

I'll quickly give it a shot after merging.

klausi (approved):
Will merge this now, thanks!

klausi:
Ah sorry, your code is correct and we cannot make distinction fixable/not fixable otherwise. All good!

tests/Drupal/Commenting/DocCommentUnitTest.1.inc
jonathan1055: It's important not to be dull, and help the maintainer enjoy the work.

jonathan1055:
OK, thanks for checking that. 
Is there any benefit/efficiency in the preg_match to using the shorthand character classes `\w\d`  for words and digits or even `alnum`, instead of listing out longhand `[a-zA-Z0-9]`?

If we do change it, post-commit, please can it be done within this PR, as I use the url  https://github.com/pfrenssen/coder/pull/130.diff within patches to utilise these fixes when the test runs only use a tagged Coder release (such as in d.o. core)

klausi:
I think `[a-zA-Z0-9]` is fine, I find it way more readable than `[\w\d]`. So no changes planned!

jonathan1055:
That suits me. I find it much more readable, and that is how my mind was thinking when I wrote it. 
I think Coder runs quite fast anyway, so shaving off micro-seconds is not a major target for us. Is that right?

klausi:
Coder runs fast enough at least :) I don't think you will be able to benchmark a performance difference when changing a regex pattern minimally like this in PHP.

PR #131, opened 2020-11-27
tests(github): Enable Github Actions testing

klausi:
See https://www.drupal.org/project/coder/issues/3185155


klausi:
Hm, does not trigger on Github, will merge and then check again.

PR #133, opened 2020-11-27
tests(github): Enable PHP 8 testing

klausi:
Checking if we can already test on PHP 8.


klausi:
I get different phpunit errors locally with PHP 8.0.0RC3, so I think we will need to upgrade phpunit.

But that would mean we have to rewrite our tests and drop older PHP 7 versions. Then we should probably make a new major version for Coder, but I would like to wait for PHPCS 4.0 before doing that.

klausi:
I think we should wait for PHP CodeSniffer 4 and then make Coder 9 for PHP 8, closing for now.

PR #135, opened 2021-01-11
Update allowed deprecation version pattern

JohnAlbin:
https://www.drupal.org/project/coder/issues/3192093

### DIFF
--- a/coder_sniffer/Drupal/Sniffs/Semantics/FunctionTriggerErrorSniff.php
+++ b/coder_sniffer/Drupal/Sniffs/Semantics/FunctionTriggerErrorSniff.php
@@ -152,13 +152,12 @@ class FunctionTriggerErrorSniff extends FunctionCall
             $phpcsFile->addError($error, $argument['start'], $sniff, [$messageText]);
         } else {
             // The text follows the basic layout. Now check that the version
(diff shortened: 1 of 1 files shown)


jonathan1055:
Thanks @JohnAlbin yes I wrote those two sniffs before the semantic versioning for projects was available. Someone else updated the DeprecatedSniff but we forgot the TriggerError one at that time.

Two things:

1. This will need additional test data to prove the new version is allowed
2. I think the use of the word "version" in the message `project:n.n.n-version[n]` is confusing. The whole thing is the version, what it really means is the 'alpha2' or 'rc1' text. Can we use "label" instead of "version" there? This would also mean we should have to change the Deprecated sniff too. But the result would be better, and now is the time to do both.

klausi:
Closing this for now due to inactivity, let me know if you want to continue or open a new pull request!

jonathan1055:
This would be useful, and is going to become more important. I will open a new PR if you would like me to pick it up. I did not before as I left it for the o.p. to progress. Or we can re-open this, as it has the history.

klausi:
If you are taking over than better make a new pull request, I think we you cannot make changes on John's branch. Thanks!

jonathan1055:
I have opened PR #153 to finish off this work

PR #141, opened 2021-07-22
Use phpcs.xml, not phpcs.xml.dist

hansfn:
Even though phpcs.xml.dist technically works - ref $defaultFiles in https://github.com/squizlabs/PHP_CodeSniffer/blob/master/src/Config.php - it is contrary to normal usage. .dist files are normally not read and is used only for reference.

If we avoid using .dist files here, we are safe if PHP_CodeSniffer ever changes it's behavior and people using Coder will be less confused - like I was.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,10 +75,10 @@ Automatically fix coding standards
     phpcbf --standard=Drupal --extensions=php,module,inc,install,test,profile,theme,css,info,txt,md,yml /file/to/drupal/example_module
 
 
-## Store settings in a phpcs.xml.dist file
+## Store settings in a phpcs.xml file
 
 In order to save and commit your PHPCS settings to Git you can use a
-phpcs.xml.dist file in your project like this:
+phpcs.xml file in your project like this:
 
 ```xml
 <?xml version="1.0" encoding="UTF-8"?>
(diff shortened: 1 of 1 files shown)


jonathan1055:
Hi @hansfn 

> it is contrary to normal usage. .dist files are normally not read and is used only for reference.

Is that really the case? Just want to check, as many contrib modules have phpcs.xml.dist, as does core.

hansfn:
Hm, maybe I missed the point - the leading sentence:

> In order to save and commit your PHPCS settings to Git you can use a phpcs.xml.dist file in your project like this

IMHO as a normal user you should create phpcs.xml to control the execution of phpcs - maybe as a copy of a default phpcs.xml.dist from some source.

However, the PHP eco system is using .dist files in a for me unfamiliar way: The .dist file is loaded if the non .dist file isn't present. I'm used to the pattern where you copy the file - like we in Drupal copy default.settings.php to settings.php

Feel free to close this PR without further action. I guess most PHP developers is used to this.

jonathan1055:
> The .dist file is loaded if the non .dist file isn't present.

Yes I think that is exactly what is intended here. You are right that it is not the same procedure as for  `default.settings.php`. The `.dist` file acts as the default, for a developer to copy and remove .dist if they want to modify it. But it is used as-is when it is the only one.

klausi:
Correct, *.dist files are great so that developers can have local modifications to test something or have a customized config.

Will not merge this, let me  know if you have any questions.

PR #145, opened 2021-12-02
3252220 wait for

alexpott:
Resolves https://www.drupal.org/project/coder/issues/3252220

PR #146, opened 2021-12-02
waitFor return check since waitForText doesn't assert

smmccabe:
Improved version of https://github.com/pfrenssen/coder/pull/145 which was initially only internal. The one has test and matches code standards

### DIFF
--- a/coder_sniffer/DrupalPractice/Sniffs/FunctionCalls/WaitForTextSniff.php
+++ b/coder_sniffer/DrupalPractice/Sniffs/FunctionCalls/WaitForTextSniff.php
@@ -1,0 +1,83 @@
+<?php
+/**
+ * \DrupalPractice\Sniffs\FunctionCalls\WaitForText
+ *
+ * @category PHP
+ * @package  PHP_CodeSniffer
+ * @link     http://pear.php.net/package/PHP_CodeSniffer
+ */
+
+namespace DrupalPractice\Sniffs\FunctionCalls;
+
+use PHP_CodeSniffer\Files\File;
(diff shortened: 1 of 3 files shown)


coder_sniffer/DrupalPractice/Sniffs/FunctionCalls/WaitForTextSniff.php
```diff
@@ -0,0 +1,83 @@
...
+
+
+    /**
+     * Returns an array of function names this test wants to listen for.
+     *
+     * @return array<string>
+     */
+    public function registerFunctionNames()
+    {
+        return ['waitForText'];
```
alexpott, line 40: Should we add more of the waitFor* methods?

alexpott:
Nice one @smmccabe - this is looking pretty good and very useful. One small question about whether to include the other waitFor* methods.

tests/DrupalPractice/FunctionCalls/WaitForTextUnitTest.php
```diff
@@ -0,0 +1,57 @@
+<?php
+/**
+ * Unit test class for the DbQuery sniff.
```
klausi, line 3: copy paste leftover?

klausi (changes requested):
Thanks, makes sense in general to me. Can we add some safe guards that we do not trigger on the function call waitForText() and that we are in a test class?

I feel like this check would be a better fit for drupalcheck, which can do real semantic analysis with phpstan https://github.com/mglaman/drupal-check

PR #147, opened 2021-12-21
Add dependency on slevomat

alexpott:
https://www.drupal.org/project/coder/issues/3010032


alexpott:
If I change the to extend 
I get the failures, including
```
Fixed version of UnusedUseStatementUnitTest.inc does not match expected version in UnusedUseStatementUnitTest.inc.fixed; the diff is
--- tests/Drupal/../../tests/Drupal/Classes/UnusedUseStatementUnitTest.inc.fixed
+++ PHP_CodeSniffer
@@ -7,6 +7,8 @@
 use Test\TraitTest;
 use Thing\DifferentName as UsedOtherName;
 use Example\MyUrlHelper;
+use MyNamespace\Depth\UnusedSameNamespace;
+/* Oh yes I do */
 use MyNamespace\Depth\SomeClass as CoreSomeClass;

 /**
@@ -54,7 +56,7 @@
   /**
    * Don't need to use classes in the same namespace.
    */
-  protected function test4(UnusedSameNamespace $x, AnotherUnusedSameNamespace $y) {
+  protected function test4(UnusedSameNamespace $x, \AnotherUnusedSameNamespace $y) {

   }

/Volumes/dev/coder/tests/Drupal/CoderSniffUnitTest.php:224
```
So I think we need to raise an issue against slevomat to fix this.

alexpott:
So the funny thing is that if I add Coder's tests/Drupal/Classes/UnusedUseStatementUnitTest.inc and use their test infra to run it through their sniff it does not break in the same way. That implies to me that the problem is on our side and we have additional sniffs that are not compatibility with their one.

alexpott:
FWIW we can add this dependency to coder and fix this in a follow-up - there are other sniff in the slevomat package we'd like to use.

composer.json
```diff
@@ -14,11 +14,12 @@
     ],
     "license": "GPL-2.0-or-later",
     "require": {
-        "php": ">=7.0.8",
+        "php": ">=7.1",
```
klausi, line 17: do we have to bump the PHP version? I would like to keep it if we don't have to drop it.

klausi (changes requested):
Thanks, but I think we should try to do something with the newly added dependency, otherwise it is of no use?

composer.json
alexpott: Yep we do... see https://github.com/slevomat/coding-standard/blob/1ace9db4f34d34692f5342f6e51fe7145472cb7b/composer.json#L18

alexpott:
@klausi 
> Thanks, but I think we should try to do something with the newly added dependency, otherwise it is of no use?
Adding the sniffs for:
https://www.drupal.org/project/drupal/issues/3222251
https://www.drupal.org/project/drupal/issues/3222769

I'm not sure what the plan is for coder with Drupal 7 / 9 / 10 compatibility... with the different PHP version requirements.

coder_sniffer/Drupal/ruleset.xml
```diff
@@ -6,6 +6,7 @@
  <arg name="encoding" value="utf-8"/>
 
  <autoload>./coder_unique_autoload_phpcs_bug_2751.php</autoload>
+ <config name="installed_paths" value="../../slevomat/coding-standard,../../vendor/slevomat/coding-standard"/>
```
alexpott: I'm not sure about this. Ideally we'd be able to include sniffs using a fully qualified classname - if they are available to the autoloader and not have to mess with PHPCS configuration at all. For the unused variable analysis we added a class that extended the `sirbrillig/phpcs-variable-analysis` but that does not feel right.

bbrala: I would probably vote agianst adding the installed_paths. Although the documentation on how to add the required sniffs manually needs updating then. (https://www.drupal.org/docs/contributed-modules/code-review-module/installing-coder-sniffer#s-register-coder-standards)

Also, shouldn't the codesniffer autoload file take care of loading the sniffs automaticly? At least when `vendor/bin/phpcs` is used? (see https://github.com/squizlabs/PHP_CodeSniffer/search?q=autoload)

alexpott: Well the problem is that the tests fail without this line. The whole way standards are loaded and PHPCS uses global configuration is a PITA. Thinking about it more I think we need to open a PHPCS issue on the best way to extend other people's coding standards which is what we want to do here.

bbrala: Phpunit runs fine locally without since the bootstrap includes the autoload.php...? Or perhaps this was because I've got the global composer plugin. I'll have an extra look.

bbrala: I have no global packages. So im kinda confused why it would fail if we remove this. It really shouldn't.

Actually, it seems the composer plugin is actually installed with the new dependency, which in turn will mean the standard is loaded?

```
$ composer why dealerdirect/phpcodesniffer-composer-installer
slevomat/coding-standard  7.0.18  requires  dealerdirect/phpcodesniffer-composer-installer (^0.6.2 || ^0.7)
```

composer.json
```diff
@@ -14,11 +14,12 @@
...
         "ext-mbstring": "*",
         "squizlabs/php_codesniffer": "^3.5.6",
         "symfony/yaml": ">=2.0.5",
-        "sirbrillig/phpcs-variable-analysis": "^2.10"
+        "sirbrillig/phpcs-variable-analysis": "^2.10",
+        "slevomat/coding-standard": "^7.0"
```
bbrala: In regards to composer 2.2, i think we need this:

```
    "config": {
        "allow-plugins": {
            "dealerdirect/phpcodesniffer-composer-installer": true
        }
    }
```

coder_sniffer/Drupal/ruleset.xml
alexpott: @bbrala nice - this all works - I guess I wonder if we should add a direct dependency on dealerdirect/phpcodesniffer-composer-installer to coder?

bbrala: Yeah i think we should since this will make it explicit and it helps in using this package.

tests/Drupal/good/good.php
```diff
@@ -1175,7 +1175,7 @@ function test6(array $names) {
 
 }
 
-list(,, $bundle) = entity_extract_ids('node', $entity);
```
klausi, line 1178: We should copy this example to bad.php so that we make sure we have a small bit of test coverage, also with the fixer.

Same for SlevomatCodingStandard.ControlStructures.RequireNullCoalesceOperator - we can add a bad.php example that is flagged (and maybe even fixed).

But that can be follow-ups, I would like to check this with our usual Coder setup in projects if it breaks anything.

I think it is also fine to drop PHP 7.0 support, has been en of life for 3 years now.

Will merge this now.

klausi (approved):

PR #148, opened 2021-12-22
Add a sniff for insecure use of unserialize().

travis-bradbury:
https://www.drupal.org/project/coder/issues/3255726

### DIFF
--- a/coder_sniffer/DrupalPractice/Sniffs/FunctionCalls/InsecureUnserializeSniff.php
+++ b/coder_sniffer/DrupalPractice/Sniffs/FunctionCalls/InsecureUnserializeSniff.php
@@ -1,0 +1,98 @@
+<?php
+/**
+ * \DrupalPractice\Sniffs\FunctionCalls\InsecureUnserializeSniff
+ *
+ * @category PHP
+ * @package  PHP_CodeSniffer
+ * @link     http://pear.php.net/package/PHP_CodeSniffer
+ */
+
+namespace DrupalPractice\Sniffs\FunctionCalls;
+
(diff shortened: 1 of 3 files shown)


coder_sniffer/DrupalPractice/Sniffs/FunctionCalls/InsecureUnserializeSniff.php
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * Record a violation of the standard.
+     *
+     * @param \PHP_CodeSniffer\Files\File $phpcsFile The file being scanned.
+     * @param int                         $position  The stack position of the violation.
+     *
+     * @return void
+     */
+    protected function fail($phpcsFile, $position) {
```
klausi: please use type hints for all function parameters

klausi (changes requested):
Cool, I think this is a great start! Can you fix the coding standards and merge in the 8.3.x branch so that we also see the test runs on PHP 8.1?

coder_sniffer/DrupalPractice/Sniffs/FunctionCalls/InsecureUnserializeSniff.php
travis-bradbury: `fail()` now has type hints:

```php
    protected function fail(File $phpcsFile, int $position)
```

`processFunctionCall()` still only has a type hint for `$phpcsFile` - the other sniffs don't have type hints of `int` for `$stackPtr`, `$openBracket`, or `$closeBracket`. Let me know if you want those to have hints too.

PR #153, opened 2022-02-19
Deprecation error messages don't allow module:n.n.n versions

jonathan1055:
D.O. issue Deprecation error messages don't allow module:n.n.n versions
https://www.drupal.org/project/coder/issues/3192093

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,2 @@
-.gitignore
 /vendor/
 /composer.lock
-.phpunit.result.cache
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
+.gitignore
 /vendor/
 /composer.lock
+.phpunit.result.cache
--- a/coder_sniffer/Drupal/Sniffs/Commenting/DeprecatedSniff.php
+++ b/coder_sniffer/Drupal/Sniffs/Commenting/DeprecatedSniff.php
@@ -169,12 +169,12 @@ class DeprecatedSniff implements Sniff
             }
         } else {
             // The text follows the basic layout. Now check that the versions
-            // match drupal:n.n.n or project:n.x-n.n or project:n.x-n.n-version[n]
-            // or project:n.n.n or project:n.n.n-version[n].
+            // match drupal:n.n.n or project:n.x-n.n or project:n.x-n.n-label[n]
+            // or project:n.n.n or project:n.n.n-label[n].
             // The text must be all lower case and numbers can be one or two digits.
             foreach (['deprecation-version' => $matches[1], 'removal-version' => $matches[2]] as $name => $version) {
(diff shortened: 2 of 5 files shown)


jonathan1055:
I have added test coverage. But given that the code fix was pushed first, what's the best way to demonstrate that the tests fail without this change? I could try re-ordering the commits, but not sure I will be allowed to do a forced push? And that is messy, re-doing the work. Any suggestions?

jonathan1055:
Rebased to re-order the commits, then pushed the first (not HEAD) with `-f` so this should fail the tests.

jonathan1055:
As expected, the updated tests fail with existing code, and we get
````
There was 1 failure:

1) Drupal\Test\Semantics\FunctionTriggerErrorUnitTest::testSniff
[LINE 41] Expected 0 warning(s) in FunctionTriggerErrorUnitTest.2.inc but found 2 warning(s). The warning(s) found were:
 
-> The deprecation-version 'my_project:1.7.0' does not match the lower-case machine-name standard: drupal:n.n.n or project:n.x-n.n (Drupal.Semantics.FunctionTriggerError.TriggerErrorVersion)

 -> The removal-version 'my_project:2.0.0' does not match the lower-case machine-name standard: drupal:n.n.n or project:n.x-n.n (Drupal.Semantics.FunctionTriggerError.TriggerErrorVersion)

[LINE 49] Expected 0 warning(s) in FunctionTriggerErrorUnitTest.2.inc but found 2 warning(s). The warning(s) found were:

 -> The deprecation-version 'my_project:2.8.0' does not match the lower-case machine-name standard: drupal:n.n.n or project:n.x-n.n (Drupal.Semantics.FunctionTriggerError.TriggerErrorVersion)

 -> The removal-version 'my_project:3.0.0' does not match the lower-case machine-name standard: drupal:n.n.n or project:n.x-n.n (Drupal.Semantics.FunctionTriggerError.TriggerErrorVersion)

**Comment by jonathan1055** [comment]
Now the fix

**Comment by jonathan1055** [comment]
Also improved comments and better grammar in the test code data

File: .gitignore [hunk 1]
```diff
@@ -1,2 +1,4 @@
+.gitignore
```
klausi: we should not ignore the gitignore itself, right?

klausi (changes requested):
Looks good to me, thanks! Can you remove the unrelated .gitignore changes? We should make a separate issue if we want to fix anything there.

.gitignore
jonathan1055: sorry I did not mean that to get in, I meant to remove the change. It was caused because I ran phpunit from within my local Coder git repository and it created that cache file, which kept getting in the way so I aded it to .gitignore. Then that introduced the change to .gitigore itself. I always assumed that each projects .gitignore is ignored.
  
Actually I do have .gitignore in my global ignore file. I'm not sure about whether its right to ignore or not ignore it per project. I always thought it was added for the convenience of developers, and it is not really part of the project.

PR #154, opened 2022-02-21
@trigger_error deprecation sniff is too strict - allow the message to start with a variable

longwave:
https://www.drupal.org/project/coder/issues/3226902

### DIFF
--- a/coder_sniffer/Drupal/Sniffs/Semantics/FunctionTriggerErrorSniff.php
+++ b/coder_sniffer/Drupal/Sniffs/Semantics/FunctionTriggerErrorSniff.php
@@ -69,6 +69,11 @@ class FunctionTriggerErrorSniff extends FunctionCall
         // Get the first argument passed to trigger_error().
         $argument = $this->getArgument(1);
 
+        // Skip variable deprecation messages.
+        if ($tokens[$argument['start']]['code'] === T_VARIABLE) {
+            return;
+        }
+
         // Extract the message text to check. If if it formed using sprintf()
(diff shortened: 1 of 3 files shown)


jonathan1055:
I can see that you had to adjust the list of warnings and errors in `FunctionTriggerErrorUnitTest.php` which is a pain, as none of those lines are affected, it's just that the line numbers all shift when new 'ok' rows are added. When I wrote the sniff and the tests, I would not have done it that way if I knew now that it would mean altering the number. It would have been better if I'd had three separate test files, one for passes, one for errors and one for warning, and then new lines could always be added at the end. 

I will see what @klausi says, but after this issus is done, I'd be happy to reorganise the test files. This wont be the last time this thing happens. I also have clashing changes on #153 so might wait for that one too.

klausi:
Thanks - merged #153 , can you resolve the conflicts here?

The number changes don't look so bad to me, I was able to understand them :)

jonathan1055:
I caused a conflict with #153 I will fix it

jonathan1055:
I don't think I can push to `longwave:relax-deprecation` so I'm not sure of the best way to fix the conflict without creating my own branch of this work and a hence a new PR. There should be a way to do it directly, it is my lack of understanding of github

klausi:
Pull requests on Github are always owned by their creator, so you will have to make your own unfortunately.

longwave:
I merged in the changes from #153.

jonathan1055:
Thanks @[longwave](https://github.com/longwave) sorry I caused you that.

PR #155, opened 2022-02-21
FunctionCommentSniff - Fatal error: str_repeat()

NickDickinsonWilde:
Fixes https://www.drupal.org/project/coder/issues/3212527

### DIFF
--- a/tests/Drupal/Commenting/FunctionCommentUnitTest.inc.fixed
+++ b/tests/Drupal/Commenting/FunctionCommentUnitTest.inc.fixed
@@ -552,3 +552,21 @@ function test38($a, $b) {
 function test39(mixed $arg) {
   return $arg;
 }
+
+/**
+ * Test multiline comments with excess spaces in the @param annotation..
+ */
+class Test40 {
+
+  /**
+   * Test method.
+   *
+   * @param string $errorMessage
+   *   It only breaks when this comment spans multiple lines but won't break if
+   *   you put it on a single line.
+   */
+  public function test41(string $errorMessage) {
+
+  }
+
+}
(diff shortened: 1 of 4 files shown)


tests/Drupal/Commenting/FunctionCommentUnitTest.inc.fixed
```diff
@@ -552,3 +552,21 @@ function test38($a, $b) {
 function test39(mixed $arg) {
   return $arg;
 }
+
+/**
+ * Test multiline comments with excess spaces in the @param annotation..
```
klausi: double dot here?

NickDickinsonWilde: accidental, fixing.

tests/Drupal/Commenting/FunctionCommentUnitTest.inc.fixed
```diff
@@ -552,3 +552,21 @@ function test38($a, $b) {
 function test39(mixed $arg) {
   return $arg;
 }
+
+/**
+ * Test multiline comments with excess spaces in the @param annotation.
```
klausi: the @ sign should also be removed here

NickDickinsonWilde: Fixed. That passed on my local as it was - I presume I mis-called phpunit somehow on that - any known easy way to mess that up?

klausi: not sure how that could go wrong - now the coding standards are failing, I feel a bit bad for you now :-|

But after that I think it should really work.

jonathan1055: @NickDickinsonWilde I expect you know, but the coding standards used in the Coder module are a bit different from the rest of Drupal. What command were you using to the check the standards? If running locally then your phpunit will probably not check any standards.

NickDickinsonWilde: ah bother. I ran phpcs from a big project on that specific path. That's embarrassing. 
Thanks!

PR #157, opened 2022-02-23
PHP attributes between docblock and function

claudiu-cristea:
See https://www.drupal.org/project/coder/issues/3250986

### DIFF
--- a/tests/Drupal/good/good.php
+++ b/tests/Drupal/good/good.php
@@ -1069,9 +1069,6 @@ class Foo implements FooInterface {
    * @return \Drupal\locale\StringInterface[]
    *   Array of objects of the class requested.
    */
-  #[
-    \ReturnTypeWillChange
-  ]
   protected function dbStringLoad(array $conditions, array $options, $class) {
     $strings = array();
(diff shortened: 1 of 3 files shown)


tests/Drupal/good/good.php
```diff
@@ -1069,6 +1069,9 @@ public function test41() {
    * @return \Drupal\locale\StringInterface[]
    *   Array of objects of the class requested.
    */
+  #[
+    \ReturnTypeWillChange
+  ]
```
klausi: this does not look normal, or is that the PHP coding standard way you write annotations?

Also please don't change existing test code, let's add new code at the end of good.php

klausi (changes requested):
Thanks, the approach makes sense to me.

Can you make the test case a bit more standalone? Also not sure what is going on with the `[]` spacing there?

tests/Drupal/good/good.php
claudiu-cristea: > this does not look normal, or is that the PHP coding standard way you write annotations?

Normally is used on a single line but, by spreading on multiple lines, I wanted to prove that works also in a non-standard way (edge case).

PR #158, opened 2022-03-01
Support advanced data types

dpi:
_Originally https://www.drupal.org/project/coder/issues/3253472_

Allow advanced type styles as implemented by [PHPStan](https://phpstan.org/writing-php-code/phpdoc-types) and [Psalm](https://psalm.dev/docs/annotating_code/type_syntax/atomic_types/), and [recognised by PHPStorm](https://blog.jetbrains.com/phpstorm/2021/07/phpstorm-2021-2-release/).

Core will be making more use of PHPStan over time. Its going to be necessary to support the types it implements.

For example array shapes:

`array{entity_type_id:string,langcode:string} `

`array<string, string>`

`array<string, SomethingMoreComplex>`

Right now to use these its a frustrating dance of ignores and in some circumstances wholesale disable/modification of sniffs.

### DIFF
--- a/tests/Drupal/Commenting/FunctionCommentUnitTest.inc
+++ b/tests/Drupal/Commenting/FunctionCommentUnitTest.inc
@@ -318,10 +318,10 @@ function test26($var) {
 /**
  * Data types with spaces.
  *
- * @param string or int $x
- *   Data types should not contain spaces.
+ * @param array<int, \stdClass> $x
+ *   We don't care that data types contain spaces.
  *
- * @return int or bool
+ * @return array<int, \stdClass>
  *   Same spaces in data type here.
  */
 function test27($x) {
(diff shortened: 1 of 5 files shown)


dpi:
Missing tests per comment on d.o

klausi:
Thanks, tests are failing with this change. Can you check? We probably have to adopt some things that were disallowed so far.

Please also merge in the 8.3.x branch, I made a phpstan fix there.

dpi:
Tried experimenting by taking the code example from `test27` and `test29` which were broken by this initial work. PHPStan doesn't complain about them, so I'm thinking PHPStan simply doesnt care if the types dont exist.

This suggests to me we shouldnt bother doing validation of the specific type in this project, so long as there is _something_ present, thats enough.

dpi:
@klausi 

I've added a bunch of tests which need to be fixed. The test are primarily examples of valid types from https://phpstan.org/writing-php-code/phpdoc-types.

I dont think it would make sense to have a giant whitelist of allowed types, perhaps there should just be minimal intelligence about the allowed characters that comprise a type. Then defer any more advanced intelligence to static analysis. Should `\Drupal\Sniffs\Commenting\FunctionCommentSniff::suggestType` remain or be reworked?

I'm gonna need some direction about reducing existing and increasing existing sniffs/suggests.

dpi:
The tests in `FunctionCommentUnitTest.inc` should mostly remain as close to samples in https://phpstan.org/writing-php-code/phpdoc-types, then we adapt from there.

tests/Drupal/Commenting/FunctionCommentUnitTest.inc
```diff
@@ -318,10 +318,10 @@ function test26($var) {
 /**
  * Data types with spaces.
  *
- * @param string or int $x
- *   Data types should not contain spaces.
+ * @param array<int, \stdClass> $x
```
klausi, line 321: Hm, can we only allow spaces after a comma? That would be more strict and should satisfy your array<> use cases.

Ideally we should still flag such broken type specifications like here.

tests/Drupal/Commenting/FunctionCommentUnitTest.inc
```diff
@@ -318,10 +318,10 @@ function test26($var) {
...
+ *   We don't care that data types contain spaces.
  *
- * @return int or bool
+ * @return array<int, \stdClass>
```
klausi, line 324: same here

tests/Drupal/Commenting/FunctionCommentUnitTest.inc
```diff
@@ -544,3 +544,293 @@ class Test40 {
...
+ * @param resource $param14
+ *   Parameter.
+ * @param void $param15
+ *   Parameter.
+ * @param object $param16
+ *   Parameter.
+ *
+ * @see https://phpstan.org/writing-php-code/phpdoc-types#basic-types
+ */
+function test41(int $param1, string $param2, mixed $param3, bool $param4, mixed $param5, mixed $param6, mixed $param7, float $param8, double $param9, mixed $param10, array $param11, iterable $param12, callable $param13, mixed $param14, mixed $param15, object $param16) {
```
klausi, line 600: $param5 type should be bool? We should not give bad examples with using mixed when this should be bool. Same for others, especially arrays.

tests/Drupal/Commenting/FunctionCommentUnitTest.inc
```diff
@@ -544,3 +544,293 @@ class Test40 {
...
+ * @param int<0, 100> $param3
+ *   Parameter.
+ * @param int<min, 100> $param4
+ *   Parameter.
+ * @param int<50, max> $param5
+ *   Parameter.
+ *
+ * @see https://phpstan.org/writing-php-code/phpdoc-types#integer-ranges
+ */
+function test42(mixed $param1, mixed $param2, mixed $param3, mixed $param4, mixed $param5) {
```
klausi, line 620: all param types should be int here.

tests/Drupal/Commenting/FunctionCommentUnitTest.inc
```diff
@@ -544,3 +544,293 @@ class Test40 {
...
+ * @param array<int, Type> $param3
+ *   Parmameter.
+ * @param non-empty-array<Type> $param4
+ *   Parmameter.
+ * @param non-empty-array<int, Type> $param5
+ *   Parmameter.
+ *
+ * @see https://phpstan.org/writing-php-code/phpdoc-types#general-arrays
+ */
+function test43(mixed $param1, mixed $param2, mixed $param3, mixed $param4, mixed $param5) {
```
klausi, line 640: all param types should be array here

tests/Drupal/Commenting/FunctionCommentUnitTest.inc
```diff
@@ -544,3 +544,293 @@ class Test40 {
...
+ * @param Collection<Type> $param2
+ *   Parameter.
+ * @param Collection<int, Type> $param3
+ *   Parameter.
+ * @param Collection|Type[] $param4
+ *   Parameter.
+ *
+ * @see https://phpstan.org/writing-php-code/phpdoc-types#iterables
+ */
+function test45(mixed $param1, mixed $param2, mixed $param3, mixed $param4) {
```
klausi, line 672: type should all by iterable?

tests/Drupal/Commenting/FunctionCommentUnitTest.inc
```diff
@@ -544,3 +544,293 @@ class Test40 {
...
+
+/**
+ * PHPStan: Union types.
+ *
+ * @param Type1|Type2 $param1
+ *   Parameter.
+ *
+ * @see https://phpstan.org/writing-php-code/phpdoc-types#union-types
+ */
+function test46(mixed $param1) {
```
klausi, line 684: type should be `Type1|Type2`

tests/Drupal/Commenting/FunctionCommentUnitTest.inc
```diff
@@ -544,3 +544,293 @@ class Test40 {
...
+
+/**
+ * PHPStan: Intersection types.
+ *
+ * @param Type1&Type2 $param1
+ *   Parameter.
+ *
+ * @see https://phpstan.org/writing-php-code/phpdoc-types#intersection-types
+ */
+function test47(mixed $param1) {
```
klausi, line 696: type should be Type1&Type2

klausi (changes requested):
Thanks, I think it makes sense to support all the PHPStan things!

You use quite a lot of "mixed" as actual type hints, but we should use all the new PHP 8 features to make sure that is supported as well. Can you make the type hints in the test cases more precise?

We should also check upstream because this sniff is copied from `PHP_CodeSniffer\Standards\Squiz\Sniffs\Commenting\FunctionCommentSniff`, maybe they already have support for similar things.

klausi:
Regarding `suggestType()`: I think this should stay, because `::$invalidTypes` contains a list of very common mistakes people make that should be flagged.

I think we have to bite the bullet and make a huge allow list of PHPStan types with a regex? Or maybe we can allow anything and only flag those simple cases? We could say: if the doc type only contains alphanumeric characters + `\` then the typehint must also only contain alphanumeric characters. That way we ensure class/interface names stay in sync between doc comment and type hint, a mistake that people often make.

Otherwise we allow really bad type strings that people not using PHPStan put in.

tests/Drupal/Commenting/FunctionCommentUnitTest.inc
dpi: I dont think this can work as things like `@return (T is int ? static : array<static>)` are possible.

I think at best types should be any non whitespace character. in the case of `@param` only one whitespace before $paramName, and only newline in the case of `@return`

dpi:
irt `mixed` everywhere I was just focussing on the docs since it doesnt really matter that there is a mismatch.

And we cant check that doc and real param match since there are many types that simply dont match eachother. Think any non core-PHP type, like `class-string` etc.

I'm hoping upstream simply has loosened these checks and we can just undo a bunch of Coder' customizations.

> I think we have to bite the bullet and make a huge allow list of PHPStan types with a regex? 

I'd really hope a whitelist isnt the solution. Maybe check with upstream? / or phpfixer

`suggestType` should still check for common mistakes, but it shouldn't always have an answer. It should give up if the type is too foreign-looking.

I dont have time for this at the moment, and I've been meaning to look at viability of `phpcsfixer`. 

This issue might be on-theme for @mglaman at the moment as there is a desire to add PHPStan types to core, which I thoroughly endorse.

klausi:
`class-string` doc comment type must always mean `string` typehint. So that would be a way of mapping for suggestType().

Apart from the `@template` case are there other examples that don't have a clear mapping like this?

tests/Drupal/Commenting/FunctionCommentUnitTest.inc
neclimdul: I was running into the same problem when trying to implement this. Not because its explicitly a bad idea, but its a bit to niave. 

That said, the original test is valid. or rather `string or int` is an invalid type hint so seems like the test shouldn't be removed.

I tried taking a stab at this issue a while back and ran out of time. I was struggling with this same check and was I think trying to make a "type checker" that could be a kinda naive lexxer and know where spaces could exist but its hard. I don't think I'd even started considering conditional types like that either.

tests/Drupal/Commenting/FunctionCommentUnitTest.inc
neclimdul: No its correctly and its explicitly a value type phpstan and psalm support. Like false it doesn't really make sense the way its used in this example but its provided for completeness so you can use it in a union type.

I've seen a bit of discussion about allowing true as a type along with false natively but I don't think its made it to an RFC since its admittedly a pretty odd case so I don't think anyone has the itch to scratch so static analysis hits like this are the only way to do it currently.

ultimike:
This would be a big +1 for me.

eileenmcnaughton:
It would be great to see this resolved - looks like there was a bunch of work done but it went cold

klausi:
Absolutely, and some parts of this are already committed now! I want to finish the PHPStan array shape support in https://www.drupal.org/project/coder/issues/3253472 (not sure how soon I will get to it, feel free to beat me to it), then I think we can call that one fixed and open new issues for more PHPStan types.

I will close this PR now as we will not merge it as is, but thanks a lot for the inspiration @dpi and others!

PR #159, opened 2022-03-14
Issue #3269399: Exclude ".min.css" files from the Drupal ruleset

joshsedl:
https://www.drupal.org/project/coder/issues/3269399

### DIFF
--- a/coder_sniffer/Drupal/ruleset.xml
+++ b/coder_sniffer/Drupal/ruleset.xml
@@ -298,6 +298,6 @@
  <exclude-pattern>*/\.hg/*</exclude-pattern>
  <exclude-pattern>*/\.bzr/*</exclude-pattern>
  
- <!-- Ignore min.css files-->
- <exclude-pattern>*.min.css</exclude-pattern>
+ <!-- Ignore minified CSS files.-->
+ <exclude-pattern>*\.min\.css$</exclude-pattern>
 </ruleset>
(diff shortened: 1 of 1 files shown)


coder_sniffer/Drupal/ruleset.xml
```diff
@@ -297,4 +297,7 @@
  <exclude-pattern>*/\.svn/*</exclude-pattern>
  <exclude-pattern>*/\.hg/*</exclude-pattern>
  <exclude-pattern>*/\.bzr/*</exclude-pattern>
+ 
+ <!-- Ignore min.css files-->
```
klausi: should be "Ignore minified CSS files. "

coder_sniffer/Drupal/ruleset.xml
```diff
@@ -297,4 +297,7 @@
...
+ <exclude-pattern>*.min.css</exclude-pattern>
```
klausi: since dots are special characters in regex we should escape them with backslashes, right? We should also add a $ at the end.

klausi (changes requested):
Thanks, just 2 minor comments, then I think we can merge this!

coder_sniffer/Drupal/ruleset.xml
joshsedl: Yes, totally makes sense, thanks!

joshsedl:
I implemented the requested changes, thanks for pointing them out @klausi!

PR #160, opened 2022-04-10
Issue #2857906 :Enforce short array syntax for Drupal 8 api docblock comments

Boegie:
https://www.drupal.org/project/coder/issues/2857906


jonathan1055:
In the [drupal issue 2857906](https://www.drupal.org/project/coder/issues/2857906) we had patch 4 which created `DocCommentArraySniff.php` and I've compared this to your new file `DocCommentLongArraySyntaxSniff.php` which I presume is a replacement for it? There are a few differences. Is it worth explaining what you had to do, given that @klausi had reviewed and made comments back then. Thanks

jonathan1055:
Sorry, its fine, we can see that you started with patch 4 above (I should have read this first, instead of just comparing the final changes).

.github/workflows/testing.yml
```diff
@@ -46,4 +46,4 @@ jobs:
           ./vendor/bin/phpcs -p --standard=tests/Drupal/phpcs-ruleset.xml tests/Drupal/good/ --ignore=tests/Drupal/good/GoodUnitTest.php
           ./vendor/bin/phpcs -p --standard=coder_sniffer/DrupalPractice tests/DrupalPractice/good/ --ignore=tests/DrupalPractice/good/GoodUnitTest.php
       - name: Run PHPStan
-        run: ./vendor/bin/phpstan analyse
+        run: ./vendor/bin/phpstan analyse
```
jonathan1055: Was this an intentional change, to remove the trailing line break?

tests/Drupal/Commenting/DocCommentLongArraySyntaxUnitTest.php
```diff
@@ -0,0 +1,54 @@
...
+     * The key of the array should represent the line number and the value
+     * should represent the number of errors that should occur on that line.
+     *
+     * @param string $testFile The name of the file being tested.
+     *
+     * @return array<int, int>
+     */
+    protected function getErrorList(string $testFile): array
+    {
+        switch ($testFile) {
```
klausi: we don't need the switch statement, right? We only have 1 test file.

coder_sniffer/Drupal/Sniffs/Commenting/DocCommentLongArraySyntaxSniff.php
```diff
@@ -0,0 +1,73 @@
...
+     */
+    public function process(File $phpcsFile, $stackPtr)
+    {
+        $tokens     = $phpcsFile->getTokens();
+        $commentEnd = $phpcsFile->findNext(T_DOC_COMMENT_CLOSE_TAG, ($stackPtr + 1));
+
+        // Look for @code annotations.
+        $codeEnd = $stackPtr;
+        do {
+            $codeStart = $phpcsFile->findNext(T_DOC_COMMENT_STRING, ($codeEnd + 1), $commentEnd, true, '@code');
```
klausi: not sure why you do an exclude find here? Can you go for `T_DOC_COMMENT_TAG` instead? That should match the tag?

coder_sniffer/Drupal/Sniffs/Commenting/DocCommentLongArraySyntaxSniff.php
```diff
@@ -0,0 +1,73 @@
...
+            if ($codeStart !== false) {
+                $codeEnd = $phpcsFile->findNext(T_DOC_COMMENT_STRING, ($codeStart + 1), $commentEnd, true, '@endcode');
```
klausi: same here

tests/Drupal/Commenting/DocCommentLongArraySyntaxUnitTest.inc.fixed
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * @file
+ * Test long array syntax in doc comments.
+ */
+
+/**
+ * Long array syntax used.
+ * @code
+ * $table = [
```
klausi: you did not implement a fixer, so the tests should actually fail because this can never correct itself?

I think we can't implement a fixer for it right now, so this file should be removed.

klausi (changes requested):
Thank you, approach makes sense to me! I have some minor remarks, please fix those and then I think we can merge.

I also think we have a bug in our test framework since this did not fail on the *.fixed file, but should not be related to this issue.

coder_sniffer/Drupal/Sniffs/Commenting/DocCommentLongArraySyntaxSniff.php
Boegie: Done

coder_sniffer/Drupal/Sniffs/Commenting/DocCommentLongArraySyntaxSniff.php
Boegie: Done

tests/Drupal/Commenting/DocCommentLongArraySyntaxUnitTest.inc.fixed
Boegie: Removed.

coder_sniffer/Drupal/Sniffs/Commenting/DocCommentLongArraySyntaxSniff.php
```diff
@@ -0,0 +1,73 @@
...
+        $codeEnd = $stackPtr;
+        do {
+            $codeStart = $phpcsFile->findNext(T_DOC_COMMENT_TAG, ($codeEnd + 1), $commentEnd, false, '@code');
+            if ($codeStart !== false) {
+                $codeEnd = $phpcsFile->findNext(T_DOC_COMMENT_TAG, ($codeStart + 1), $commentEnd, false, '@endcode');
+                if ($codeEnd !== false) {
+                    // Check for long array syntax use inside this @code annotation.
+                    for ($i = ($codeStart + 1); $i < $codeEnd; $i++) {
+                        if (preg_match('/\barray\s*\(/', $tokens[$i]['content']) === 1) {
+                            $error = 'Long array syntax used in doc comment code annotation';
```
klausi, line 62: maybe better "Long array syntax must not be used in doc comment code annotations'

klausi (approved):
Thanks, looks good. The message could be improved, but I can do that after the merge.

PR #161, opened 2022-04-11
Issue #3271952: Validate function names are prefixed with module name.

vbouchet31:
https://www.drupal.org/project/coder/issues/3271952

### DIFF
--- a/coder_sniffer/Drupal/Sniffs/NamingConventions/ValidFunctionNameSniff.php
+++ b/coder_sniffer/Drupal/Sniffs/NamingConventions/ValidFunctionNameSniff.php
@@ -28,6 +28,16 @@ use PHP_CodeSniffer\Util\Common;
 class ValidFunctionNameSniff extends CamelCapsFunctionNameSniff
 {
 
+    /**
+     * A list of function prefixes which may not respect naming convention.
+     *
+     * @var string[]
+     */
+    protected $allowedFunctionPrefixes = [
+        'template_preprocess',
+        'theme',
+    ];
+
 
     /**
      * Processes the tokens within the scope.
@@ -106,7 +116,7 @@ class ValidFunctionNameSniff extends CamelCapsFunctionNameSniff
         $isApiFile     = substr($phpcsFile->getFilename(), -8) === '.api.php';
         $isHookExample = substr($functionName, 0, 5) === 'hook_';
         if ($isApiFile === true && $isHookExample === true) {
-            // Ignore for examaple hook_ENTITY_TYPE_insert() functions in .api.php
+            // Ignore for example hook_ENTITY_TYPE_insert() functions in .api.php
             // files.
             return;
         }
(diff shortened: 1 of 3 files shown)


coder_sniffer/Drupal/Sniffs/NamingConventions/ValidFunctionNameSniff.php
```diff
@@ -28,6 +28,13 @@
 class ValidFunctionNameSniff extends CamelCapsFunctionNameSniff
 {
 
+    /**
+     * A list of function prefixes which may not respect naming convention.
+     *
+     * @var string[]
+     */
+    protected $functionPrefixes = ['template_preprocess'];
```
klausi: for Drupal 7 we also need to allow the prefix "theme" (and "phptemplate", probably not that relevant anymore. See https://git.drupalcode.org/project/pareviewsh/-/blob/7.x-1.x/pareview.sh#L285

Let's add a test case for this in a *.module file for the ValidFunctionNameUnitTest

coder_sniffer/Drupal/Sniffs/NamingConventions/ValidFunctionNameSniff.php
```diff
@@ -121,6 +128,25 @@ protected function processTokenOutsideScope(File $phpcsFile, $stackPtr)
...
+                if (substr($functionName, 0, strlen($functionPrefix)) === $functionPrefix) {
+                    return;
+                }
+            }
+
+            $moduleName = reset($nameParts);
+            if (preg_match("/^_?$moduleName\_.+/", $functionName) === 0) {
+                $error = 'All functions defined in a module file must be prefixed with the module\'s name, found "%s".';
+                $data  = [$functionName];
+                $phpcsFile->addError($error, $stackPtr, 'InvalidName', $data);
```
klausi: let's use a new error code here instead of 'InvalidName', so that people can make targeted exceptions for this in their configurations. I would suggest to use "InvalidPrefix".

tests/Drupal/bad/bad.module
```diff
@@ -49,3 +49,10 @@ function bad_menu() {
   );
   return $items;
 }
+
+/**
+ * This function does not respect the naming convention as it is not prefixed.
+ */
+function not_respecting_naming_convention_with_module_prefix() {
```
klausi: Let's move this to the ValidFunctionNameUnitTest directory and make a *.module file there.

coder_sniffer/Drupal/Sniffs/NamingConventions/ValidFunctionNameSniff.php
```diff
@@ -121,6 +128,25 @@ protected function processTokenOutsideScope(File $phpcsFile, $stackPtr)
             $phpcsFile->addError($error, $stackPtr, 'InvalidName', $data);
         }
 
+        $nameParts     = explode('.', basename($phpcsFile->getFilename()));
```
klausi: let's use the same approach as with *.api.php files above with substr().

coder_sniffer/Drupal/Sniffs/NamingConventions/ValidFunctionNameSniff.php
klausi: Let's rename this to "allowedFunctionPrefixes"

coder_sniffer/Drupal/Sniffs/NamingConventions/ValidFunctionNameSniff.php
```diff
@@ -121,6 +128,25 @@ protected function processTokenOutsideScope(File $phpcsFile, $stackPtr)
...
+            // Check if the function prefix is allowed to not respect standard.
+            foreach ($this->functionPrefixes as $functionPrefix) {
+                if (substr($functionName, 0, strlen($functionPrefix)) === $functionPrefix) {
+                    return;
+                }
+            }
+
+            $moduleName = reset($nameParts);
+            if (preg_match("/^_?$moduleName\_.+/", $functionName) === 0) {
+                $error = 'All functions defined in a module file must be prefixed with the module\'s name, found "%s".';
```
klausi: We should give an example what we expect like "All functions defined in a module file must be prefixed with the module\'s name, found "%s" but expected "%s"". For the expected name we can just prefix the function name with `$moduleName . '_'`

klausi (changes requested):
Thanks, the approach make sense! I have some minor comments to improve test coverage and output, then I think we can go ahead with this.

vbouchet31:
Thanks for the review @klausi . I updated the PR.

PR #162, opened 2022-05-11
Issue #3081260:  Fix ternary operator indent validation

JParkinson1991:
Addresses: https://www.drupal.org/project/coder/issues/3081260

Previously `ArraySniff` would enforce:

````php
$array = [
  ('ternary' == true)
  ? 'no'
  : 'indent',
]
````

Pull request will enforce **single additional level of indentation** for ternary operator components.

Tested against indexed, associative and multi-dimensional arrays.

Examples:

````php

// Good
$array = [
  ('ternary' == true)
    ? 'single'
    : 'indent',
]

// Bad
$array = [
  ('ternary' == true)
  ? 'no'
  : 'indent',
  ('ternary' == true)
      ? 'too much'
      : 'indent',
]

````

Thanks

### CODE REVIEW

**Review by klausi — CHANGES_REQUESTED** [review]
Thanks, makes sense to me. Can you fix the coding standards in the PHPCS changes? That is failing right now.

The test cases are super useful and seem correct, nice work!

**Comment by JParkinson1991** [comment]
Fixed the coding standard errors in the recent commits. Ready to be retested.

Note that there may be an issue with the `Squiz.Operators.ComparisonOperatorUsage.ImplicitTrue` sniff.

The below snippet seems to be triggering PHPCS errors

````php
if (in_array($tokens[$newLineStart]['code'], $increasedNestingTokens, true)) {
   ...
}
````
````
---------------------------------------------------------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
---------------------------------------------------------------------------------------------------------------------------------
 267 | ERROR | Implicit true comparisons prohibited; use === TRUE instead
     |       | (Squiz.Operators.ComparisonOperatorUsage.ImplicitTrue)
---------------------------------------------------------------------------------------------------------------------------------
````

To fix I've removed the in_array() check result to a distinct variable and checked that variables value when determining wether extra indentation to be applied. Now reading:

````php
$extraIndentRequired = (in_array($tokens[$newLineStart]['code'], $extraIndentTokens, true));

if ($extraIndentRequired === true) {
    ...
}
````

File: coder_sniffer/Drupal/Sniffs/Arrays/ArraySniff.php [hunk 1]
```diff
@@ -254,10 +254,22 @@ public function process(File $phpcsFile, $stackPtr)
...
+            // Expects: '->', '?', ':'.
+            $extraIndentTokens = [
+                T_OBJECT_OPERATOR,
+                T_INLINE_THEN,
+                T_INLINE_ELSE,
+            ];
+
+            // Determine whether extra indentation required for the next array line.
+            // True where the line starts with one of the defined extra indent tokens.
+            $extraIndentRequired = (in_array($tokens[$newLineStart]['code'], $extraIndentTokens, true));
```
klausi, line 267: the extra true parameter should not be necessary, but also does not hurt.

The coding standard error was related to the if(), where you need to compare the result of in_array() with a bool.

But this style is also perfectly fine.

klausi (approved):

coder_sniffer/Drupal/Sniffs/Arrays/ArraySniff.php
JParkinson1991: That makes more sense, i must have misread the sniff error.

Assume the following would not be flagged as error:

````php
if (in_array($tokens[$newLineStart]['code'], $increasedNestingTokens, true) === true) {
   ...
}
````

Either way, thanks for merging!

PR #163, opened 2022-05-19
@param tags must use fully-qualified class names, even when the full namespace is given without the leading backslash.

Boegie:
See https://www.drupal.org/project/coder/issues/3281300


klausi:
Thank you! As explained in the issue please fix this in the existing sniff instead.

PR #164, opened 2022-05-24
Issue #3123282: Do not require @var tag if a property has typehint

claudiu-cristea:


pfrenssen:
Thanks for working on this! Linking the drupal.org issue: https://www.drupal.org/project/coder/issues/3123282, and the coding standards issue: https://www.drupal.org/project/coding_standards/issues/3238192#new

pfrenssen:
In `VariableCommentUnitTest.inc`, let's update this example because the documentation is no longer current.

```
  /**
   * A typed property should still contain a variable comment.
   */
  public ?Bar $bar;
```

claudiu-cristea:
@pfrenssen good catch. Done!

pfrenssen (approved):

coder_sniffer/Drupal/Sniffs/Commenting/VariableCommentSniff.php
```diff
@@ -111,6 +111,15 @@ public function processMemberVar(File $phpcsFile, $stackPtr)
 
         // The @var tag is the only one we require.
         if ($foundVar === null) {
+            // If there's an inline type argument then you may omit the @var
+            // comment.
+            $propertyName       = $phpcsFile->findNext([T_VARIABLE, T_CLASS, T_INTERFACE], ($commentEnd + 1));
```
klausi, line 116: This should not be needed, because the property name pointer is already in `$stackPtr`, right?

Then I think we can simplify this logic: if the non-whitespace token before $stackPtr is T_STRING then we know that is a type token and we can return.

klausi (changes requested):
Thank you, this change makes sense to me! Can we simplify the logic?

coder_sniffer/Drupal/Sniffs/Commenting/VariableCommentSniff.php
claudiu-cristea: @klausi, my knowledge here are extremely limited, so, no, I don’t get your point, sorry. Actually, it wasn’t me writing this code, I’ve only added test coverage

klausi: I can check, should not be hard to rewrite this.

klausi:
Finally merged that one after some new phpstan updates, thank you!

PR #174, opened 2022-09-19
Incomplete attempt at @todo capitalization enforcement.

adamzimmermann:
This might end up being closed.

Discussion here:
- https://git.drupalcode.org/project/drupal/-/merge_requests/504#note_119707

Issue:
- https://www.drupal.org/i/3180696

### DIFF
--- a/coder_sniffer/Drupal/Sniffs/Commenting/TodoCommentSniff.php
+++ b/coder_sniffer/Drupal/Sniffs/Commenting/TodoCommentSniff.php
@@ -142,7 +142,7 @@ class TodoCommentSniff implements Sniff
             (?-i)              # Reset to case-sensitive
             (?!                # Start another non-consuming look-ahead, this time negative
               @todo\s          # It has to match lower-case @todo followed by one space
-              (?!-|:)\S        # and then any non-space except "-" or ":".
+              (?!-|:)[A-Z]     # and then any capital letter except "-" or ":".
             )/m';
 
         if ((bool) preg_match($expression, $comment, $matches) === true) {
--- a/tests/Drupal/Commenting/TodoCommentUnitTest.inc
+++ b/tests/Drupal/Commenting/TodoCommentUnitTest.inc
@@ -7,9 +7,9 @@
  * These are valid examples.
  *
  * @todo Valid.
- * @todo valid with lower-case first letter
- * @todo $can start with a $
- * @todo \also with backslash
+ * @todo Valid
+ * @todo Multiple words are valid.
+ * @todo Multiple words are valid
  *
  * These are all incorrect but can be fixed automatically.
  *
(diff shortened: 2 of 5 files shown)


jonathan1055:
On the d.o. issue you said
>My vote is to take the gains we have and move forward to begin seeing the benefits.

Yes I agree. If I can help with anything, let me know

klausi:
Thanks, can you make an issue in the Coder issue queue? Then we can track credit and work there aside from the Drupal core issue.

tests/Drupal/Commenting/TodoCommentUnitTest.inc
```diff
@@ -7,9 +7,9 @@
  * These are valid examples.
  *
  * @todo Valid.
- * @todo valid with lower-case first letter
- * @todo $can start with a $
```
klausi, line 11: this should be valid, one could document a variable name like "@todo $entity could be NULL here, fix this later in foo_fcuntion()."

tests/Drupal/Commenting/TodoCommentUnitTest.inc
```diff
@@ -7,9 +7,9 @@
...
- * @todo \also with backslash
```
klausi, line 12: Same here, backslash should be valid. "@todo \Drupal should not be used here, implement dependency injection later".

coder_sniffer/Drupal/Sniffs/Commenting/TodoCommentSniff.php
```diff
@@ -142,7 +142,7 @@ private function checkTodoFormat(File $phpcsFile, int $stackPtr, string $comment
             (?-i)              # Reset to case-sensitive
             (?!                # Start another non-consuming look-ahead, this time negative
               @todo\s          # It has to match lower-case @todo followed by one space
-              (?!-|:)\S        # and then any non-space except "-" or ":".
+              (?!-|:)[A-Z]     # and then any capital letter except "-" or ":".
```
klausi, line 145: this is tricky. We want to warn about lower case characters, but allow everything else. Even lower case characters should be allowed in some cases, for example `@todo some_function() is not a good name, refactor it to a better name`.

I think we have code somewhere else in Coder that checks for such things in some comment standard, we should check that.

jonathan1055:
Yes, in response to @klausi's points above, I think the logic could be reverted to allow  `any non-space except "-" or ":"` as we had before. Then we give a separate warning if the first letter is a lower-case alpha and provide the fix to convert to upper case.

adamzimmermann:
I have created an issue to facilitate further discussion as requested.

- https://www.drupal.org/project/coder/issues/3314583

I tried to summarize the consensus there, but if others could chime in with their understanding of the requirements that would be great. Then I can begin working on implementing it.

PR #176, opened 2023-02-01
stdclass is auto-corrected to object but it should not be 

mxr576:
Original MR: https://git.drupalcode.org/project/coder/-/merge_requests/20
D.o issue: https://www.drupal.org/i/3313246


klausi:
Fixed in #188

PR #180, opened 2023-02-27
Avoid misleading message when inheritdoc is missing the curly braces

jonathan1055:
d.o. issue https://www.drupal.org/project/coder/issues/2904801

### DIFF
--- a/tests/Drupal/Commenting/DocCommentUnitTest.inc.fixed
+++ b/tests/Drupal/Commenting/DocCommentUnitTest.inc.fixed
@@ -137,11 +137,11 @@ function test14(array $matches, array $sub_key, $to) {
 function test15_this_inheritdoc_is_correct();
 
 /**
- * @inheritdoc
+ * {@inheritdoc}
  */
 function test16_lower_case_fail_needs_braces();
 
 /**
- * {@inheritdoc}
+ * {@inheritDoc}
  */
 function test17_camel_case_fail_needs_braces();
--- a/tests/Drupal/Commenting/DocCommentUnitTest.inc.fixed
+++ b/tests/Drupal/Commenting/DocCommentUnitTest.inc.fixed
@@ -130,3 +130,18 @@ function test13() {
 function test14(array $matches, array $sub_key, $to) {
 
 }
+
+/**
+ * {@inheritdoc}
+ */
+function test15_this_inheritdoc_is_correct();
+
+/**
+ * @inheritdoc
+ */
+function test16_lower_case_fail_needs_braces();
+
+/**
+ * {@inheritdoc}
+ */
+function test17_camel_case_fail_needs_braces();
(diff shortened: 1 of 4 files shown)


jonathan1055:
Intentionaly incorrect 'fixed' file and no update to expected lines, to demonstrate that the tests run and fail. Corrected in next commit.

tests/Drupal/Commenting/DocCommentUnitTest.inc.fixed
```diff
@@ -130,3 +130,18 @@ function test13() {
...
+ */
+function test15_this_inheritdoc_is_correct();
+
+/**
+ * {@inheritdoc}
+ */
+function test16_lower_case_fail_needs_braces();
+
+/**
+ * {@inheritDoc}
```
klausi, line 145: The official docs say to use `{@inheritdoc}` with a lower case `d` https://www.drupal.org/docs/develop/standards/php/api-documentation-and-comment-standards#inheritdoc . I'm a bit on the fence if the fixer should produce wrong output. Ideally it would lower-case the `d` in the same go?

Don't think this is a blocker, if you agree please fix.

tests/Drupal/Commenting/DocCommentUnitTest.php
```diff
@@ -33,6 +33,8 @@ protected function getErrorList(string $testFile): array
                 66  => 1,
                 100 => 4,
                 101 => 1,
+                130 => 1,
```
klausi: wrong line number, the error is on 131. Can we use the pointer to the tag instead?

klausi (changes requested):
Thanks, approach makes sense. I have 2 minor comments that should be quick to resolve.

tests/Drupal/Commenting/DocCommentUnitTest.inc.fixed
jonathan1055: I think I recall some wider discussion somewhere about which to use. You are right that the drupal standards page shows lowercase in the example, but it does not actually state that it must be lower case. 

I also found https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/issues/5141 which talks about the camel case version being preferrable for PHPstan, and in https://docs.phpdoc.org/guide/guides/inheritance.html#inheritance all the examples are camel case. 

I'd be happy to change the fixer if Coder itself had a sniff to highlight that inheritDoc is wrong and should be changed to inheritdoc. But until that rule is decided, and a sniff created, I think it is best to leave the fixer to use what ever is currently in the file.

tests/Drupal/Commenting/DocCommentUnitTest.php
jonathan1055: Good call, have changed this now.

tests/Drupal/Commenting/DocCommentUnitTest.inc.fixed
klausi: Thanks for researching! Then I agree with you that we should not mess with this for now, as upper case `D` might turn out to be even preferred.

jonathan1055: I have created a small follow-up PR #185 to add a camel case example to the test file. This is in prep for a future sniff which may enforce one or the other. Having the four examples close in the file will make it simpler to adjust the fixed file and expected lines, rather than add the fourth case later, after other test examples may have been added at the end.

PR #182, opened 2023-03-01
Allow T_STRING argument type to allow passing constants

### LINKED ISSUE #181: Allow passing constants to t() translate function to prevent 'Only string literals should be passed to t() where possible' CS warning
It's a clone of the original issue in Drupal.org https://www.drupal.org/project/coder/issues/3326197

# Problem/Motivation

Now, when we try to pass a constant to the `t()` function, the code sniffer shows the warning:
`Only string literals should be passed to t() where possible`

But, as for me, using a constant is also ok, it's the same as a string literal.
Using constants is very convenient when we have several places to insert the same text: with constants, we can just put it once in a constant, and reuse it in all needed places, especially in different classes.

So, maybe we can find some way to allow this?

The current workaround to dismiss the warning is to use `// phpcs:ignore` or `@codingStandardsIgnoreStart` / `@codingStandardsIgnoreEnd` tricks around the `t()` call.

# Steps to reproduce

1. Get the code like this:
```php
define("ALIAS_RULES_DESCRIPTION", "Alias should start with /");
$description = $this->t(ALIAS_RULES_DESCRIPTION);
```

2. Run the code sniffer and see the warning.

# Proposed resolution

Try to detect if there is a constant passed, and if so - dismiss the warning.

I've found a way via allowing also T_STRING argument type, via this change:
```php
        if (
          $tokens[$argument['start']]['code'] !== T_CONSTANT_ENCAPSED_STRING
          && $tokens[$argument['start']]['code'] !== T_STRING
        ) {
```

It works well, and if I'm trying to pass a string with a inline variable like `"foo $bar"` - the argument type is `T_DOUBLE_QUOTED_STRING`, so it still works well for such case.

Could anyone check if this is a suitable fix, or it can also allow some "bad things" to pass?

MurzNN:
Issue: https://www.drupal.org/project/coder/issues/3326197

### DIFF
--- a/coder_sniffer/Drupal/Sniffs/Semantics/FunctionTSniff.php
+++ b/coder_sniffer/Drupal/Sniffs/Semantics/FunctionTSniff.php
@@ -78,6 +78,7 @@ class FunctionTSniff extends FunctionCall
         if (
           $tokens[$argument['start']]['code'] !== T_CONSTANT_ENCAPSED_STRING
           && $tokens[$argument['start']]['code'] !== T_STRING
+          && $tokens[$argument['start']]['code'] !== 'PHPCS_T_SELF'
         ) {
             // Not a translatable string literal.
             $warning = 'Only string literals should be passed to t() where possible';
--- a/coder_sniffer/Drupal/Sniffs/Semantics/FunctionTSniff.php
+++ b/coder_sniffer/Drupal/Sniffs/Semantics/FunctionTSniff.php
@@ -77,7 +77,7 @@ class FunctionTSniff extends FunctionCall
 
         if ($tokens[$argument['start']]['code'] !== T_CONSTANT_ENCAPSED_STRING
             && $tokens[$argument['start']]['code'] !== T_STRING
-            && $tokens[$argument['start']]['code'] !== 'PHPCS_T_SELF'
+            && $tokens[$argument['start']]['code'] !== T_SELF
         ) {
             // Not a translatable string literal.
             $warning = 'Only string literals should be passed to t() where possible';
--- a/coder_sniffer/Drupal/Sniffs/Semantics/FunctionTSniff.php
+++ b/coder_sniffer/Drupal/Sniffs/Semantics/FunctionTSniff.php
@@ -75,7 +75,10 @@ class FunctionTSniff extends FunctionCall
             return;
         }
 
-        if ($tokens[$argument['start']]['code'] !== T_CONSTANT_ENCAPSED_STRING) {
+        if (
+          $tokens[$argument['start']]['code'] !== T_CONSTANT_ENCAPSED_STRING
+          && $tokens[$argument['start']]['code'] !== T_STRING
+        ) {
             // Not a translatable string literal.
             $warning = 'Only string literals should be passed to t() where possible';
             $phpcsFile->addWarning($warning, $argument['start'], 'NotLiteralString');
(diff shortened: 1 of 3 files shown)


klausi:
Thanks, triggered the tests now which are failing. Can you add a test case for FunctionTSniff? There should be a test file named similarly.

MurzNN:
Still can't reproduce the failing tests locally, I've run the tests locally and all of them are passing:
```
$ git branch
* 3326197-allow-constants-in-t
  8.3.x
$ ./vendor/bin/phpunit 
PHPUnit 8.5.33 by Sebastian Bergmann and contributors.

...............................................................  63 / 102 ( 61%)
.......................................                         102 / 102 (100%)

Time: 1.3 minutes, Memory: 26.00 MB

OK (102 tests, 14 assertions)
```

MurzNN:
Added one more case with self constants.

tests/Drupal/Semantics/FunctionTUnitTest.inc
```diff
@@ -21,6 +21,8 @@ $x = t('Test') . ' (disabled)';
 
 class Test {
 
+  const FOO2 = 'bar2';
```
klausi: we don't need this line, Coder does not parse the PHP code semantics to check if it exists.

klausi (changes requested):
Thanks, coding standards are failing now. Can you also fix them?

tests/Drupal/Semantics/FunctionTUnitTest.inc
MurzNN: Fixed, thanks!

MurzNN:
Fixed code standards.

coder_sniffer/Drupal/Sniffs/Semantics/FunctionTSniff.php
```diff
@@ -75,7 +75,10 @@ public function processFunctionCall(
             return;
         }
 
-        if ($tokens[$argument['start']]['code'] !== T_CONSTANT_ENCAPSED_STRING) {
+        if ($tokens[$argument['start']]['code'] !== T_CONSTANT_ENCAPSED_STRING
+            && $tokens[$argument['start']]['code'] !== T_STRING
+            && $tokens[$argument['start']]['code'] !== 'PHPCS_T_SELF'
```
klausi: why are your referring to the string here? This can be T_SELF, which is defined in PHPCS, right?

tests/Drupal/Semantics/FunctionTUnitTest.inc
```diff
@@ -34,3 +34,8 @@ class Test {
   }
 
 }
+
+// Using constants in t() should be allowed.
+$x = t(FOO);
+$x = t(Test::FOO);
+$x = t(self::FOO);
```
klausi: Ah sorry, now I understand your previous changes.

I think the self test case is not correct here, this should be called within a class. Self outside of classes is invalid, does this even parse as T_SELF? Best move this test in the class above in a new test method and refer to the class constant there.

klausi (changes requested):
Sorry, overlooked some other things.

tests/Drupal/Semantics/FunctionTUnitTest.inc
MurzNN: Thanks for the idea, moved.

coder_sniffer/Drupal/Sniffs/Semantics/FunctionTSniff.php
MurzNN: Right, fixed.

MurzNN:
Fixed, please take a look.

PR #185, opened 2023-03-05
Add camelCase inheritDoc example to test file

jonathan1055:
This is a follow-up to #180 to simply add a camel case example of inheritDoc which does not currently need to be fixed and will pass. Having all four examples here together in the test file will make it easier to test a future sniff which may enforce either lowercase or camelcase. Do this now, before any more lines are added to the .inc and .inc.fixed files.


klausi (approved):
Sure, thanks!

jonathan1055:
Thank you.

PR #186, opened 2023-03-07
Ensure module dependencies are an array in .info.yml

jonathan1055:
d.o. issue https://www.drupal.org/project/coder/issues/3166470
This is a new PR following the initial work on #127 which was closed due to inactivity.
I have improved the sniff and corrected the test file.


tests/Drupal/InfoFiles/drupal8/dependencies_array.info.yml
```diff
@@ -0,0 +1,4 @@
+core: 8.x
+name: Dependencies Array Test
+dependencies :
```
klausi, line 3: any reason for the space here?

klausi (approved):
Looks good, thanks!

tests/Drupal/InfoFiles/drupal8/dependencies_array.info.yml
jonathan1055: Yes I added it to prove that [this preg_match](https://github.com/pfrenssen/coder/blob/8.3.x/coder_sniffer/Drupal/Sniffs/InfoFiles/DependenciesArraySniff.php#L69) would cope if the user left space(s) before the colon. I did not want to just check for the plain string 'dependencies' as that could pick up something else. But also checking for 'dependencies:' would miss the line if there were spaces.

PR #187, opened 2023-03-07
Add sniff to check and fix Use statement order

jonathan1055:
d.o. issue https://www.drupal.org/project/coder/issues/3310013
The first commit adds the new sniff whcih is the work of cburschka who provided a patch on the drupal issue.
The other commits, to fix the 'good' file and the four '.inc.fixed' files demonstrate that the new sniff works, has test coverage, and that the fixer works. 

Question - @klausi  do you want an extra specific test to show the sniff working? The fact that I had to change the good.php file and four .fixed files shows that it does work, and these checks will continue verify that the sniff continues to work and does not get broken by something else in future.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,6 +19,7 @@
         "dealerdirect/phpcodesniffer-composer-installer": "^0.7.1 || ^1.0.0",
         "sirbrillig/phpcs-variable-analysis": "^2.11.7",
         "slevomat/coding-standard": "^7.0 || ^8.0",
+        "cweagans/composer-patches": "*",
         "squizlabs/php_codesniffer": "^3.7.1",
         "symfony/yaml": ">=3.4.0"
     },
@@ -30,9 +31,17 @@
     },
     "config": {
         "allow-plugins": {
+            "cweagans/composer-patches": true,
             "dealerdirect/phpcodesniffer-composer-installer": true
         },
         "sort-packages": true
+    },
+    "extra": {
+        "patches": {
+            "slevomat/coding-standard": {
+                "Do not fix multi-part use stmt": "https://patch-diff.githubusercontent.com/raw/slevomat/coding-standard/pull/1551.diff"
+            }
+        }
     },
     "require-dev": {
         "phpstan/phpstan": "^1.7.12",
--- a/composer.json
+++ b/composer.json
@@ -14,12 +14,11 @@
     ],
     "license": "GPL-2.0-or-later",
     "require": {
-        "php": ">=7.1",
+        "php": ">=7.2",
         "ext-mbstring": "*",
         "dealerdirect/phpcodesniffer-composer-installer": "^0.7.1 || ^1.0.0",
         "sirbrillig/phpcs-variable-analysis": "^2.11.7",
-        "slevomat/coding-standard": "^7.0 || ^8.0",
-        "cweagans/composer-patches": "*",
+        "slevomat/coding-standard": "^8.11",
         "squizlabs/php_codesniffer": "^3.7.1",
         "symfony/yaml": ">=3.4.0"
     },
@@ -35,13 +34,6 @@
             "dealerdirect/phpcodesniffer-composer-installer": true
         },
         "sort-packages": true
-    },
-    "extra": {
-        "patches": {
-            "slevomat/coding-standard": {
-                "Do not fix multi-part use stmt": "https://patch-diff.githubusercontent.com/raw/slevomat/coding-standard/pull/1551.diff"
-            }
-        }
     },
     "require-dev": {
         "phpstan/phpstan": "^1.7.12",
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,6 @@
     },
     "config": {
         "allow-plugins": {
-            "cweagans/composer-patches": true,
             "dealerdirect/phpcodesniffer-composer-installer": true
         },
         "sort-packages": true
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -7,7 +7,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-versions: ['7.1', '7.2', '7.3', '7.4', '8.0', '8.1']
+        php-versions: ['7.2', '7.3', '7.4', '8.0', '8.1']
         phpstan: ['0']
         # We only need to run PHPStan once on the latest PHP version.
         include:
--- a/coder_sniffer/Drupal/ruleset.xml
+++ b/coder_sniffer/Drupal/ruleset.xml
@@ -121,6 +121,7 @@
  <rule ref="PSR2.Namespaces.UseDeclaration" />
 
  <rule ref="SlevomatCodingStandard.ControlStructures.RequireNullCoalesceOperator" />
+ <rule ref="SlevomatCodingStandard.Namespaces.AlphabeticallySortedUses" />
  <rule ref="SlevomatCodingStandard.PHP.ShortList" />
 
  <rule ref="Squiz.Arrays.ArrayDeclaration" />
--- a/tests/Drupal/Classes/ClassCreateInstanceUnitTest.inc.fixed
+++ b/tests/Drupal/Classes/ClassCreateInstanceUnitTest.inc.fixed
@@ -1,7 +1,7 @@
 <?php
 
-use Vendor\DateTools\DateInterval;
 use Vendor\DateTools;
+use Vendor\DateTools\DateInterval;
 
 $x = array(new Foo(), array());
 $y = new Foo();
--- a/tests/Drupal/Classes/FullyQualifiedNamespaceUnitTest.inc.fixed
+++ b/tests/Drupal/Classes/FullyQualifiedNamespaceUnitTest.inc.fixed
@@ -5,12 +5,12 @@
  * Example.
  */
 
-use Test\MultiLineSecond;
-use Test\Foo;
-use Test\NotUsed;
-use Test\Bar;
 use Test\Alias as TestAlias;
+use Test\Bar;
+use Test\Foo;
 use Test\MultiLine as MultiLineAlias;
+use Test\MultiLineSecond;
+use Test\NotUsed;
 
 /**
  * Example.
--- a/tests/Drupal/Classes/UnusedUseStatementUnitTest.inc.fixed
+++ b/tests/Drupal/Classes/UnusedUseStatementUnitTest.inc.fixed
@@ -2,12 +2,12 @@
 
 namespace MyNamespace\Depth;
 
-use Thing\NotUsed;
-use Thing\Fail\ActuallyUsed;
-use Test\TraitTest;
-use Thing\DifferentName as UsedOtherName;
 use Example\MyUrlHelper;
 use MyNamespace\Depth\SomeClass as CoreSomeClass;
+use Test\TraitTest;
+use Thing\DifferentName as UsedOtherName;
+use Thing\Fail\ActuallyUsed;
+use Thing\NotUsed;
 
 /**
  * Bla.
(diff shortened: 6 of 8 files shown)


klausi (changes requested):
Thanks a lot! I think we need at least one explicit test file for this new sniff to isolate our tests better and be future-proof.

The most important question for me: are we 100% sure this ordering algorithm is the same as PHPStorm's? It would be super bad if this contradicts PHPStorm in any way. I'm not a PHPStorm user, so would appreciate another person to verify this.

jonathan1055:
Yes, that's a good question. I'm not a PHPStorm user either. There were more followers on the Coding Standards issue https://www.drupal.org/project/coding_standards/issues/1624564 so I will try to direct some of them to contribute here or on the Coder issue. I am fairly sure the the slevomat order is further away from phpstorm than what we are proposing here, so we definitely do not want to reuse that one.

jonathan1055:
On https://www.drupal.org/project/coding_standards/issues/1624564#comment-13094505 Gogowitsch said _"PhpStorm will re-order use statements to be alphabetically. This implicitly keeps the vendors grouped. The order of vendors is just alphabetically."_

But I will try to verify this.

jonathan1055:
Following the discussion on d.o. issue I have removed the new coder sniff and added in  SlevomatCodingStandard.Namespaces.AlphabeticallySortedUses to the ruleset.xml file

jonathan1055:
That's interesting, the slevomat fixer make two other changes to the files, not just the sort order. 

Firstly we have
````
1) Drupal\Test\Classes\FullyQualifiedNamespaceUnitTest::testSniff
Fixed version of FullyQualifiedNamespaceUnitTest.inc does not match expected version in
FullyQualifiedNamespaceUnitTest.inc.fixed; the diff is
--- tests/Drupal/Classes/FullyQualifiedNamespaceUnitTest.inc.fixed
+++ PHP_CodeSniffer
@@ -8,7 +8,7 @@
 use Test\Alias as TestAlias;
 use Test\Bar;
 use Test\Foo;
-use Test\MultiLine as MultiLineAlias;
+use Test\MultiLine;
 use Test\MultiLineSecond;
 use Test\NotUsed;
 
@@ -69,7 +69,7 @@
   /**
    * Description.
    */
-  public function test8(MultiLineAlias $multiLine, MultiLineSecond $multiLineSecond) {
+  public function test8(MultiLine $multiLine, MultiLineSecond $multiLineSecond) {
 ````
I don't know if this is due to it being in a multi line use statement in the before file, or whether its because the actual alias `MultiLineAlias` is not used in the file, so the fixer deems it unnecessary, drops it and renames back to the original.

The second fail is
````
2) Drupal\Test\Classes\UseGlobalClassUnitTest::testSniff
Fixed version of UseGlobalClassUnitTest.inc does not match expected version in UseGlobalClassUnitTest.inc.fixed; the diff is
--- tests/Drupal/../../tests/Drupal/Classes/UseGlobalClassUnitTest.inc.fixed
+++ PHP_CodeSniffer
@@ -52,7 +52,7 @@
    * Description.
    */
   public function test5() {
-    $test = new \MultiLineSecond();
+    $test = new MultiLineSecond();
   }
 
   /**
````
This is probably correct, I'm not sure, but it was not expected from the addition of this sniff.

But I am sure we can find a good way forward, and use slevotmat, and not add the custom sniff to coder

**Comment by klausi** [comment]
Agreed that we should move forward with the Slevomat sniff.

The fixer seems a bit eager, but I think we can use a same second class name so that the aliasing is actually needed?

The second fail seems wrong, since our coding standards say that global classes must not be declared with use statements and have the `\` when invoked.

**Comment by jonathan1055** [comment]
The fixer is definitely eager, and appears to be making changes that are out of scope. When the use statements are in the correct order and there is no sorting to be done it does nothing. The following file is considered correct, and remains unchanged:
```
use Test\Alias as TestAlias;
use Test\MultiLine as MultiLineAlias,
    Test\MultiLineSecond;
```
However, if you move the first line to the end, so that we have
```
use Test\MultiLine as MultiLineAlias,
    Test\MultiLineSecond;
use Test\Alias as TestAlias;
```
then the order is incorrect, the fixer jumps into action, and the resulting file is
```
use Test\Alias as TestAlias;
use Test\MultiLine as MultiLineAlias;
```
The second part of the multi-line alias statement gets deleted completely! Yes, this alias was not used - none of them are used in this file, but the only sniff being executed was the one to order the statements, not delete them. But probably in a real situation this might be ok, as it makes the file cleaner and easier to read if an alias is not required.

I could investigate if there any configuration options for the slevomat sniff so that we can reduce what it fixes. I will also make the change you suggest so that the alias is needed.

**Comment by jonathan1055** [comment]
Last commit solved the problem of the alias being deleted. Now we just need to solve why the new fixer _does not_ want to add the backslash. The .inc file has 
```
public function test5() {
    $test = new MultiLineSecond();
  }
```
and the original fixed file adds the `\` 
```
public function test5() {
    $test = new \MultiLineSecond();
  }
```
But during the tests now  this function is not changed and does not get the `\` added. Is it due to the order that the sniffs are executed? i.e. does our original fixer (which ever that one is) add the `\` then the slevomat fixer removes it again? Or is it that the Slevomat fixer runs first, so that Coders fixer does not detect the problem? If so, it might be that the Coder sniff needs looking at. 

Just some random ideas I had whilst writing up this comment. I will follow up with proper investigation.

**Comment by jonathan1055** [comment]
I've done some more checking, and in summary 
- The order that the sniff fixers are executed _does_ make a difference to the outcome. This is not ideal, of course, as we do not have any way to effect the sequence of sniffs within one execution of phpcs, do we?
- I also think there is a bug in the slevomat fixer that incorrectly drops use statements.
 
The details: Looking at just the two sniffs `Drupal.Classes.UseGlobalClass` and `SlevomatCodingStandard.Namespaces.AlphabeticallySortedUses` and running them individually on our test file `tests/Drupal/Classes/UseGlobalClassUnitTest.inc`

When the drupal fixer is run on its own it reports and fixes five redundant use statements. In doing so it does a lot of work: 

1. Removes `use Test;` and replaces all `Test` with `\Test`
2. Removes `use Test1 as Alias;` and replaces all `Alias` with `\Test1`
3. Removes both within the double statement `use MultiLine as MultiLineAlias, MultiLineSecond;` and replaces `MultiLineAlias` with `\MultiLine` and `MultiLineSecond` with `\MultiLineSecond`
4. Removes `MultiLineSecond2` because it is not used at all
 
So this sniff adds the \ to `$test = new \MultiLineSecond();` that we want in our fixed file.

If the slevomat fixer is run _after_ these have been done, it just reports on the wrong order and fixes it, nothing more. This means that our .fixed file is OK.

But, if the slevomat fixer is run first, it also does a lot more than just reordering the use statements. It removes the second parts of the multiline statements `MultiLineSecond` and `MultiLineSecond2`. But it does not do anything else. This leaves the `$test = new MultiLineSecond();` without any \ but also no Use statement for it. 

Then if the Drupal fixer is run afer the Slevomat one, it only reports three redundant use statements because the slevomat fixer has removed two of the original five redundant. It fixes these three and does the necessary replacements and adds \ where needed. But because the Slevomat fixer removed the use statement for `MultiLineSecond` the Drupal fixer does not do anything with this, and hence does not add the required \ to the test5 function, and we get left with the unchanged `$test = new MultiLineSecond();`. This does not match the .fixed file.

I need to do some more testing on the Slevomat sniff, but it does seem that it is dropping use statements that then cause problems later.

**Comment by jonathan1055** [comment]
Here is an example of the problem, using this test file:
```
<?php
use yourproject\A,
    myproject\B;
use anotherproject\C;
```
If the single sniff `SlevomatCodingStandard.Namespaces.AlphabeticallySortedUses` is run to fix the file, the result is
```
<?php
use anotherproject\C;
use yourproject\A;
```
The second part of the multiline `use` statement is lost. I know that the Drupal standard does not want multiline use statements, but if they are valid php then the sniff should not remove it. I think this is an unwanted by-product of the sorting. If the first parts of the use statement are already in proper sequence then nothing is done, and the multuline statement remains intact.

@klausi do you think this a bug that I should report on the Slevomat issue queue?

**Comment by klausi** [comment]
Yep, definitely a bug in Slevomat, please file an issue there!

**Comment by jonathan1055** [comment]
I debugged my version of the slevomat sniff, and found exactly where it just takes the first part of the statement and does not check if there is any more within the statement before the next semicolon. The problem is in slevomat/coding-standard/SlevomatCodingStandard/Helpers/UseStatementHelper.php

I was starting to post the issue, and went to get a link to the problem area, and discovered that this has actually been found and fixed 9 months ago. The last commit on this file says "we do not support group use statements". 
https://github.com/slevomat/coding-standard/commit/c5cf9b5c1a44b8abba1279a199dcac206d1266c3

The test runs here at php7.1 use Slevomat 7, but from php7.2 onwards it uses Slevomat 8.8 so it should run OK. It was fixed in 8.0.1 apparently https://github.com/slevomat/coding-standard/releases/tag/8.0.1

I will update my local version and see what it actually does.

**Comment by jonathan1055** [comment]
Ah, I've just realised that "group use declaration" is not the same as "multiline declaration". So this is still a bug, in slevomat 8.8 and I will report it. (tomorrow)

**Comment by jonathan1055** [comment]
Slevomat bug reported https://github.com/slevomat/coding-standard/issues/1547
Pull request https://github.com/slevomat/coding-standard/pull/1551
I have now used a patch from the slevomat pr to patch Coder here, and the tests all pass green :)
Force-pushed my branch, to remove all the previous unrequired commits

**Comment by jonathan1055** [comment]
I am now confused! The tests all pass, but the log has
```
- Applying patches for slevomat/coding-standard
    https://patch-diff.githubusercontent.com/raw/slevomat/coding-standard/pull/1551.diff (Do not fix multi-part use stmt)
   Could not apply patch! Skipping. The error was: Cannot apply patch https://patch-diff.githubusercontent.com/raw/slevomat/coding-standard/pull/1551.diff
```
So if the patch was not applied I would expect some test failures like we had before.

**Comment by jonathan1055** [comment]
I have just verified on anther test branch in my coder github that removing the patch command makes the tests fail as expected. Adding it back, and we get test passes but also that same message saying that the patch cannot be applied. I think the patch must be being applied, and attempted a second time to give that bogus message. It is very unsatisfactory not to understand what is going on.

**Comment by jonathan1055** [comment]
I added a step to display the top of the file that is being patched. As expected (shown on my other testing branch) the slevomat sniff file is patched as required, even though it shows the `"Could not apply patch! Skipping."` error message. Without the patch, the file is unchanged as expected and the tests fail. 

This is ready for review from the Coder point of view, but obviously we can't commit until Slevomat have made a new release. 8.10.0 was released 2 days ago, and they make regular releases so I am just waiting for some responses on the issue https://github.com/slevomat/coding-standard/issues/1547 and pull request https://github.com/slevomat/coding-standard/pull/1551

**Comment by jonathan1055** [comment]
I have worked out why we get the `"Could not apply patch! Skipping."` error even though the file does get patched correctly. The PR .diff file has changes for four files, the actual sniff `SlevomatCodingStandard/Sniffs/Namespaces/AlphabeticallySortedUsesSniff.php` plus three test files (one changed and two new). However, when composer installs Slevomat we do not get the Slevomet test files (don't need them in a Coder installation). So the cweagans patching process works fine on the sniff file, but fails on the test files. It is the failure to patch the slevomet test files which causes the overall `"Could not apply patch"` error, even though the sniff file (the only one we are interested in) does get patched. So that explains it.

File: composer.json [hunk 1]
```diff
@@ -30,6 +30,7 @@
     },
     "config": {
         "allow-plugins": {
+            "cweagans/composer-patches": true,
```
klausi: we don't need this line and should remove it to avoid confusion

klausi (approved):
Thanks, will cleanup composer.json and then we can merge.

composer.json
jonathan1055: sorry, yes I intended to remove that line.

PR #188, opened 2023-04-02
fix(FunctionComment): stdclass is auto-corrected to object but it should not be

klausi:
Issue https://www.drupal.org/project/coder/issues/3313246

Continued form #176


mxr576:
Thanks, the original PR was stuck on my todo list

PR #199, opened 2023-04-30
refactor(FullyQualifiedNamespace): Replace sniff in favor of a sniff from Slevomat

klausi:
Issue: https://www.drupal.org/project/coder/issues/3214374

The goal would be to delete our FullyQualifiedNamespace sniff and switch to using the one from Slevomat.

Was not able to complete this yet because of a bug in Slevomat https://github.com/slevomat/coding-standard/issues/1564


jonathan1055:
You've probably realised this, but the failing tests for standards is because the new file `tests/Drupal/good/GoodNamespace.php` is not excluded when checking Coder's own files:
```
    <!-- Test files that should not be checked. -->
    <exclude-pattern>tests/*\.(inc|css|js|api\.php|tpl\.php)$</exclude-pattern>
    <exclude-pattern>tests/*/(good|bad)\.php$</exclude-pattern>
    <exclude-pattern>tests/*/drupal(6|7|8)/*\.php$</exclude-pattern>
```
The patterns above seem to exclude nearly all files in the tests/good/ folder but not everything. Am I right in thinking that all files in the tests/good folder should follow _Drupal_ standards, and hence they are excluded because Coder uses a different customised set for it's own files? If so, we could exclude everything from tests/good in one statement, rather than try to cater for most files over the three patterns. 

Sorry if this is wrong, I've not looked in great detail at what is _not_ excluded from the above three.

klausi:
Yep, this is not ready for review yet, was just uploading my work in progress here.

jonathan1055:
Sorry, yes of course. But I've done some learning in the meantime, I wanted to see if we could exclude everything in `/good` _except_ the GoodUnitTest.php but I could not get a neagative pattern to work, so the easiest thing would be to add GoodNamespace into the `tests/*/(good|GoodNamespace|bad)\.php$` pattern

klausi:
Created upstream bug report at https://github.com/slevomat/coding-standard/issues/1570

klausi:
That bug was resolved, yay!

Now some test fails remain where Slevomat gets confused by multiple use statements, it somehow thinks one is a constant.

Not sure if we should simply drop this test case from Coder and not care about it anymore, or if we should try to improve the Slevomat sniff.

klausi:
Reported the constant confusion upstream: https://github.com/slevomat/coding-standard/issues/1581

klausi:
Giving up on this approach, as grouped use statements are still broken in Slevomat.

I'm fixing our existing sniff instead in #252 and #253 .

jonathan1055:
Yes that's a good plan.

PR #202, opened 2023-06-10
Regression in 8.3.19 for variadic function parameters

Boegie:
Should resolve https://www.drupal.org/project/coder/issues/3365993

### DIFF
--- a/coder_sniffer/Drupal/Sniffs/Commenting/FunctionCommentSniff.php
+++ b/coder_sniffer/Drupal/Sniffs/Commenting/FunctionCommentSniff.php
@@ -845,21 +845,23 @@ class FunctionCommentSniff implements Sniff
         // Missing parameters only apply to methods and not function because on
         // functions it is allowed to leave out param comments for form constructors
         // for example.
-        // It is also allowed to ommit pram tags completely, in which case we don't
+        // It is also allowed to omit param tags completely, in which case we don't
         // throw errors. Only throw errors if param comments exists but are
         // incomplete on class methods.
         if ($tokens[$stackPtr]['level'] > 0 && empty($foundParams) === false) {
             foreach ($realParams as $realParam) {
                 $realParamKeyName = $realParam['name'];
                 if (in_array($realParamKeyName, $foundParams) === false
-                    && ($realParam['pass_by_reference'] === true
-                    && in_array("&$realParamKeyName", $foundParams) === true) === false
+                    && (($realParam['pass_by_reference'] === true
+                    && in_array("&$realParamKeyName", $foundParams) === true)
+                    || ($realParam['variable_length'] === true
+                    && in_array("...$realParamKeyName", $foundParams) === true)) === false
                 ) {
                     $error = 'Parameter %s is not described in comment';
                     $phpcsFile->addError($error, $commentStart, 'ParamMissingDefinition', [$realParam['name']]);
                 }
             }
-        }
+        }//end if
 
     }//end processParams()
 
--- a/tests/Drupal/Commenting/FunctionCommentUnitTest.inc
+++ b/tests/Drupal/Commenting/FunctionCommentUnitTest.inc
@@ -916,3 +916,23 @@ function test_return_literal_string(): string {
 function test_intersection_types(Foo&Bar $a): Foo&Bar {
   return new Xyz();
 }
+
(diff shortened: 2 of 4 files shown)


jonathan1055:
I guess Coder was lacking some test coverge? This could have been discovered when the original commit was done. I'm surprised that it was missed because @klausi is normally very good on test coverage for changes.

Boegie:
Indeed, I had to add a test case for variadic parameters in a function myself.

I was surprised myself, usually there are no/always never regressions in Coder.
What I _think_ happened is that because the changes in this release aren't specifically about varidadics, klausi himself assumed that test coverage would be there.
_(n=1, YMMV, etc)_

klausi:
Ha, very sweet of you both! I'm also a minimalist and try not to overcomplicate changes to not get stuck and make incremental progress ... I think regressions like this can happen and are ok, as that only surfaced 2 places in the large Drupal core code base.

Thanks a lot for showing me the forward testing of Drupal core, this could be something for running in Coder as well to catch regressions earlier. If we can define that as CI task that is allowed to fail it could be helpful.

I think we are missing a bit of test coverage here, we should also assert that we still get an error if the param description is missing.

```php
  /**
   * Variadic parameters without description are not ok.
   *
   * @param string[]|string ...$classes
   *
   * @return array
   *   The return array.
   */
  public function testVariadicParameterNoDescription(...$classes): array {
    return [];
  }
```

Can you add that?

Boegie:
>  I think regressions like this can happen and are ok, as that only surfaced 2 places in the large Drupal core code base.

Fully agreed: Omelette, egs, only 2 places, etc.

> Thanks a lot for showing me the forward testing of Drupal core, this could be something for running in Coder as well to catch regressions earlier. If we can define that as CI task that is allowed to fail it could be helpful.

`mglaman/phpstan-drupal` started doing something (kinda) similar, maybe this helps?
https://github.com/mglaman/phpstan-drupal/issues/418

>I think we are missing a bit of test coverage here, we should also assert that we still get an error if the param description is missing.

Agreed and added.

jonathan1055:
Coder has another new regression, quite major, and will impact every project that defines constants with a `define( )` statement. 
d.o. issue https://www.drupal.org/project/coder/issues/3369978
gitlab issue #204   
I have fixed it and added test coverage - see #205

But I was wondering if we can make any progress on the discussion above about forward testing on Core or a big (and clean) module or project with the latest Coder dev? This error could have been found before committing.

Boegie:
Nice work @jonathan1055 

>  forward testing on Core or a big (and clean) module or project with the latest Coder dev? This error could have been found before committing.

In this case we would have only discovered it if did both, since Core wasn't affected by this regression (the `updateded deps` daily QA run https://www.drupal.org/node/3060/qa would have caught it).

So we really need both _**and**_ we really need it.

I'll try to whip something up this weekend if no one beats me to it.

jonathan1055:
You are right that Core won't be affected _yet_ because it does not yet include `Drupal.Semantics.ConstantName` in the sniffs it is checking. But many modules will be affected, particularly those that run gitlab-ci.yml as that immediately uses the new Coder release.

Boegie:
Yeah, wasn't exactly too clear in my previous message, it is indeed _yet_.
Testing against both is a necessity as it seems.

Any contrib modules spring to mind?

jonathan1055:
Let's create an issue on drupal.org Coder queue and copy some of the comments above, then we can have the details stored in the place where others will see and contribute.

jonathan1055:
https://www.drupal.org/project/coder/issues/3374864 and https://github.com/pfrenssen/coder/pull/207

jonathan1055:
> Any contrib modules spring to mind?

It needs to be a module that is kept up to date and clean of coding standards faults so that we don't spend too much time checking up on false positives. I maintain [Scheduler](https://www.drupal.org/project/scheduler) and keep it passing PHPCS - it only has a couple of customisations in [phpcs.xml](https://git.drupalcode.org/project/scheduler/-/blob/2.x/phpcs.xml#L24) so it's fairly 'vanilla'.

Boegie:
I tried Token, as being the most used contrib module. Turned out it was not PHPCS passing, so I stopped there.

Scheduler seems as good a candidate as any, I'll have a stab on that today.

PR #206, opened 2023-07-11
Array without key is allowed

jonathan1055:
Fixes https://www.drupal.org/project/coder/issues/3355778

Arrays which start with a key and then have elements without a key should be allowed. 

First add array without key to demonstrate the problem. Test should fail.


jonathan1055:
As intended the first test fails with
````
1) Drupal\Test\good\GoodUnitTest::testSniff
[LINE 170] Expected 0 error(s) in good.php but found 1 error(s). The error(s) found were:
 -> No key specified for array entry; first entry specifies key (Squiz.Arrays.ArrayDeclaration.NoKeySpecified)
````

klausi:
Closing this as discussed in the issue.

PR #207, opened 2023-07-16
Test for regressions against Drupal core

Boegie:
See https://www.drupal.org/project/coder/issues/3374864

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -51,3 +51,47 @@ jobs:
           ./vendor/bin/phpcs -p --standard=coder_sniffer/DrupalPractice tests/DrupalPractice/good/ --ignore=tests/DrupalPractice/good/GoodUnitTest.php
       - name: Run PHPStan
         run: if [[ ${{ matrix.phpstan }} == "1" ]]; then ./vendor/bin/phpstan analyse; fi
+
+  drupal_core_check:
+    needs:
+      - testing
+    continue-on-error: true
+    runs-on: "ubuntu-latest"
+    name: "Regression test with Drupal coding standards against Drupal core HEAD"
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v3"
+      - name: "Set the version alias for self"
+        run: |
+          if [ "${{ github.event_name }}" == 'pull_request' ]; then
+            echo "VERSION_ALIAS=dev-"${{ github.sha }}"" >> $GITHUB_OUTPUT
+          else
+            echo "VERSION_ALIAS=dev-"${{ github.ref_name }}"" >> $GITHUB_OUTPUT
+          fi
+        id: branch_alias
+      - name: "Install PHP 8.2"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          coverage: "none"
+          php-version: "8.2"
+          tools: composer:v2
+          extensions: dom, curl, libxml, mbstring, zip, pdo, mysql, pdo_mysql, gd, apcu
+      - name: "Checkout Drupal core HEAD"
+        run: |
+          cd ${{ runner.temp }}
+          git clone https://git.drupalcode.org/project/drupal.git
+          cd drupal
+          composer config repositories.0 composer https://packages.drupal.org/8
+          composer config repositories.1 path $GITHUB_WORKSPACE
+      - name: "Install Drupal core dependencies"
+        uses: "ramsey/composer-install@v2"
+        with:
+          working-directory: "${{ runner.temp }}/drupal"
+      - name: "Require drupal/coder"
+        run: |
+          cd ${{ runner.temp }}/drupal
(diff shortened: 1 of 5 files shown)


Boegie:
So now we know, as expected/wanted, the current HEAD of `drupal/coder` returns no errors when running the Drupal coding standards against Drupal core HEAD.

We also want to know:

- If the test runs correctly, not only when creating a PR but also when doing "just" a commit.
- if the test fails on a recent regression.

Let's revert "Fix regression in 8.3.19 for variadic function parameters (#3365993)" to test both,

Boegie:
Good enough for me, let's revert the revert to have only the addition of the new GitHub CI step in this PR,

jonathan1055:
NIce work. One question about the failed test
test against core

Boegie:
Nice catch.

Should be fixed now.

Fail test showing the fail:
image

Including failed mail:
image

jonathan1055:
Excellent.

jonathan1055:
This will only detect regressions in the sniffs that core has enabled, so there could be things that still get missed. It's good to see that in recent months there has been a push to get more sniffs passing and enabled in core. Here's the current list of things being worked on https://www.drupal.org/project/drupal/issues/2571965

jonathan1055:
For example, the regression in [constantName.constantStart](https://www.drupal.org/project/coder/issues/3369978) would not have been discovered because Core has not enabled this rule yet.

Boegie:
> This will only detect regressions in the sniffs that core has enabled, so there could be things that still get missed.

AFAIK, that's exactly why new stuff goes into new rules and it should be a core queue issue to fix new failures on the new rule before enabling it (in the same issue) in `core/phpcs.xml.dist`.

The regression test that's proposed in this issue is just that, testing regressions in existing rules.
I personally see no part-to-play for Coder to test new rules on Drupal core, but n=1, so open for suggestions on this one.

jonathan1055:
Yes I'm totally with you on all that. I didn't mean to sound like I was taking anything away from your work - this is a superb addition to Coder's reliability and integrity.

Boegie:
Oh, I wasn't offended/whatever in any way/shape/form, nor do I (strongly) oppose the opinion coder should test for failures on new rules.
_Blantatly playing my non-native-language card_

.github/workflows/testing.yml
```diff
@@ -51,3 +51,46 @@ jobs:
...
+        uses: "shivammathur/setup-php@v2"
+        with:
+          coverage: "none"
+          php-version: "8.2"
+          tools: composer:v2
+          extensions: dom, curl, libxml, mbstring, zip, pdo, mysql, pdo_mysql, gd, apcu
+      - name: "Checkout Drupal core HEAD"
+        run: |
+          cd ${{ runner.temp }}
+          git clone https://git.drupalcode.org/project/drupal.git
```
klausi, line 81: we should use the github git mirror for drupal core for better speed. We should also use the checkout action.

Or is that not good because we don't have a good default git branch on github?

.github/workflows/testing.yml
```diff
@@ -51,3 +51,46 @@ jobs:
...
+          cd drupal
+          composer config repositories.0 composer https://packages.drupal.org/8
+          composer config repositories.1 path $GITHUB_WORKSPACE
+      - name: "Install Drupal core dependencies"
```
klausi, line 85: we don't need the drupal core dependencies at all, we just want to check the code of drupal core.

I'm thinking if we can integrate this whole job into the PHP 8.2 testing run further above? Like we do with PHPStan.

We clone Drupal core and run the local phpcs with drupal core's phpcs config. That should be all that is needed?

klausi (changes requested):
Thanks, can we simplify this to be part of the normal PHP 8.2 test run?

.github/workflows/testing.yml
Boegie: > I'm thinking if we can integrate this whole job into the PHP 8.2 testing run further above? Like we do with PHPStan.

Wouldn't that clash with the the important green icon?
Wouldn't we be able to distinguish a fail in coder's code and a regression on core?

Anyway, this issue is getting way to complicated for lil ol' me, so I won't be persuing any chances.

klausi: Well, this pull request already does interfere with the important green icon as you demonstrated if I understand it correctly? Which is fine - if there is a regression in Drupal core we should know about it. Then there is a small chance that it is a false positive and we have to modify this Github action to work around it - which we also accept.

If we have different steps in the action we can see where it failed - either at Coders tests or at the Drupal core regression test.

No worries about the complication, anybody else can pick this up and we continue at a later point. Thanks for the contribution, nice demo in any case!

jonathan1055: ??? what happened? This was going so well. I thought we were all in agreement (mostly) and that this was nearing being merged. It was such a good addition to the Coder's stability.

Boegie: Not much happenned, I assume?

There's a demo, things can be changed in the code, just not by me, important green icons is something beyond my comprehension/paygrade :)

klausi:
Continuing this work now in #211

PR #218, opened 2024-01-03
Add sniff for unsilenced deprecation warnings

Boegie:
Resolves https://www.drupal.org/project/coder/issues/3412078

Although this could quite easily be integrated in `\Drupal\Sniffs\Semantics\FunctionTriggerErrorSniff`, I've opted for a separate rule.
Main reason being that there are two known unsilenced deprecations already.
Adding to an existing an enabled rule would break Drupal CI.


coder_sniffer/Drupal/Sniffs/Semantics/FunctionUnsilencedDeprecationSniff.php
```diff
@@ -0,0 +1,77 @@
...
+
+        $tokens = $phpcsFile->getTokens();
+
+        // If no second argument then quit.
+        if ($this->getArgument(2) === false) {
+            return;
+        }
+
+        // Only check deprecation messages.
+        if (strcasecmp($tokens[$this->getArgument(2)['start']]['content'], 'E_USER_DEPRECATED') !== 0) {
```
klausi: Instead of calling $this->getArgument(2) again, can we put this in a variable?

klausi (commented):
Approach makes sense!

Ideally we would also add a fixer for this to insert the `@`.

coder_sniffer/Drupal/Sniffs/Semantics/FunctionUnsilencedDeprecationSniff.php
```diff
@@ -0,0 +1,77 @@
...
+use PHP_CodeSniffer\Files\File;
+
+/**
+ * Checks that the trigger_error deprecation is silenced by a preceding '@'.
+ *
+ * @category PHP
+ * @package  PHP_CodeSniffer
+ * @link     http://pear.php.net/package/PHP_CodeSniffer
+ */
+class FunctionUnsilencedDeprecationSniff extends FunctionCall
```
klausi: Can we shorten this to just `UnsilencedDeprecationSniff`?

PR #230, opened 2024-06-26
Tags support is not enabled. Enable the "Yaml::PARSE_CUSTOM_TAGS" flag

FlorentTorregrosa:
PR with fix + tests

PR with only tests: https://github.com/pfrenssen/coder/pull/231


klausi:
Thanks!

klausi:
I also checked that this works with the minimum YAML 3.4.0 version we have in composer.

PR #232, opened 2024-07-19
Plugin discovery annotations with long translatable strings are allowed.

pfrenssen:
Fixes https://www.drupal.org/project/coder/issues/3462678


pfrenssen:
Oh this already works! Great, never mind :)

PR #233, opened 2024-08-03
style(cspell): Fix cspell spelling

klausi:
Issue: https://www.drupal.org/project/coder/issues/3465767


coder_sniffer/Drupal/Sniffs/CSS/ClassDefinitionNameSpacingSniff.php
```diff
@@ -98,8 +98,8 @@ public function process(File $phpcsFile, $stackPtr)
                 && in_array($tokens[($i - 1)]['code'], [T_WHITESPACE, T_COMMA]) === false
             ) {
                 $error = 'Selectors must be on a single line';
-                // cspell:ignore SeletorSingleLine
-                $fix   = $phpcsFile->addFixableError($error, $i, 'SeletorSingleLine');
+                // Cspell:ignore SeletorSingleLine .
+                $fix = $phpcsFile->addFixableError($error, $i, 'SeletorSingleLine');
```
jonathan1055: I guess the capital start letter and ending period were added to satisfy the normal comment format. But this should not be necessary, and it looks odd. I have used this comment style for `cspell:ignore` and not had to make that compromise.

klausi: Yeah, this is not nice. For Coder we are checking with the PHPCS coding standard, which uses a sniff from Squiz. That one that does not have the cspell exception.

But now that I think about it - we could replace the Squiz Sniff with the Drupal sniff when checking in Coder, that should work.

jonathan1055: Did you make any progress on the above?  I had the same problems in Gitlab Templates pipelines. `Drupal.Commenting.InlineComment.InvalidEndChar` does not respect if the comment starts with `cspell`

jonathan1055: I have just seen https://www.drupal.org/project/coder/issues/3424914 which I think deals with this problem

klausi: Fixed that and updated this PR.

PR #236, opened 2024-09-01
Add SlevomatCodingStandard.TypeHints.NullableTypeForNullDefaultValue

andypost:
Ref https://www.drupal.org/project/coder/issues/3471146#comment-15752633


klausi:
Thanks, I will add a small test case to bad.php in a follow-up.

PR #237, opened 2024-09-27
Enforce method to not have docblock

claudiu-cristea:
See https://www.drupal.org/project/coder/issues/3477407


coder_sniffer/Drupal/Sniffs/Commenting/FunctionCommentSniff.php
```diff
@@ -94,13 +101,24 @@ public function process(File $phpcsFile, $stackPtr)
...
 
-        // Constructor methods are exempt from requiring a docblock.
-        // @see https://www.drupal.org/project/coder/issues/3400560.
         $methodName = $phpcsFile->getDeclarationName($stackPtr);
-        if ($methodName === '__construct'
-            && $tokens[$commentEnd]['code'] !== T_DOC_COMMENT_CLOSE_TAG
+        if ($tokens[$commentEnd]['code'] !== T_DOC_COMMENT_CLOSE_TAG
             && $tokens[$commentEnd]['code'] !== T_COMMENT
         ) {
+            if ($methodName === '__construct') {
```
klausi, line 108: we must not treat constructors in a special way. Comments on constructors are good and we need to allow them.

klausi (changes requested):
thanks - can you introduce this in a way that existing tests stay as is?

PR #239, opened 2024-10-25
Adds better detection of DI classes to global sniff.

### LINKED ISSUE #238: Detection of DI on GlobalDrupalSniff is too loose.
Currently the following must be true to catch `\Drupal::` calls in DI classes:

1. Class doesn't extend anything OR Doesn't extend one of those defined as base classes
2. Is not a service itself
3. Doesn't implement `ContainerInjectionInterface`

So this omits a lot of legit cases of plugins as they commonly will use `ContainerFactoryPluginInterface`

b-sharpe:
Fixes #238


klausi (changes requested):
Thanks, makes sense!

Please add 2 test cases for the 2 new class names so that we have coverage for them.

batkor:
Hi.
nice MR. 
I get similar error
image

rules not detect bad code on my custom webform plugin.

Thanks!

PR #241, opened 2024-11-29
Add PHP 8.4 CI run

andypost:
TBD


klausi:
Thanks, instead of adding another special run we can move up PHP 8.4. Doing that in #242

PR #243, opened 2024-12-29
Add attributes check to ScopeIndentSniff

nlighteneddesign:


tests/Drupal/bad/bad.php
```diff
@@ -838,6 +838,26 @@ function test28() {
 enum aBad_NameForAnEnum {
 }
 
+/**
+ * Test PHP attributes.
+ */
+class TestPhpAttributes {
```
klausi: please move this test to ScopeIndentUnitTest.inc and adapt the fixed file as well as the list of errors in ScopeIndentUnitTest.php.

klausi (changes requested):
Thanks, please move the test.

tests/Drupal/bad/bad.php
nlighteneddesign: I think I addressed both of these!

klausi:
Thanks, tests are failing because the error lines are not updated. But I can do that in a quick follow-up.

PR #260, opened 2025-01-13
feat(ValidHookName): Add sniff for hook name in PHP attributes

RuZniki:
Issue: https://www.drupal.org/project/coder/issues/3499291

### DIFF
--- a/coder_sniffer/Drupal/Sniffs/Attributes/ValidHookNameSniff.php
+++ b/coder_sniffer/Drupal/Sniffs/Attributes/ValidHookNameSniff.php
@@ -22,13 +22,6 @@ use PHP_CodeSniffer\Sniffs\Sniff;
 class ValidHookNameSniff implements Sniff
 {
 
-    /**
-     * List of hooks that should not be fixed.
-     *
-     * @var string[]
-     */
-    public array $hookExceptions = ['hook_info'];
-
 
     /**
      * Returns an array of tokens this test wants to listen for.
@@ -69,9 +62,7 @@ class ValidHookNameSniff implements Sniff
                 // Remove outer quotes.
                 $hookNameValue = trim($tokens[$hookName]['content'], '"\'');
 
-                if (in_array($hookNameValue, $this->hookExceptions) === false
-                    && strpos($hookNameValue, 'hook_') === 0
-                ) {
+                if (strpos($hookNameValue, 'hook_') === 0) {
                     $fix = $phpcsFile->addFixableWarning("Hook name should not start with 'hook_'. Hook name used: $hookNameValue", $hookName, 'HookPrefix');
                     if ($fix === true && strlen($hookNameValue) > 5) {
                         // Remove "hook_" prefix.
--- a/coder_sniffer/Drupal/Sniffs/Attributes/ValidHookNameSniff.php
+++ b/coder_sniffer/Drupal/Sniffs/Attributes/ValidHookNameSniff.php
@@ -1,13 +1,13 @@
 <?php
 /**
- * \Drupal\Sniffs\NamingConventions\ValidAttributeHookNameSniff.
+ * \Drupal\Sniffs\Attribute\ValidHookNameSniff.
  *
  * @category PHP
  * @package  PHP_CodeSniffer
  * @link     http://pear.php.net/package/PHP_CodeSniffer
  */
 
-namespace Drupal\Sniffs\NamingConventions;
+namespace Drupal\Sniffs\Attributes;
 
 use PHP_CodeSniffer\Files\File;
 use PHP_CodeSniffer\Sniffs\Sniff;
@@ -19,7 +19,7 @@ use PHP_CodeSniffer\Sniffs\Sniff;
  * @package  PHP_CodeSniffer
  * @link     http://pear.php.net/package/PHP_CodeSniffer
  */
-class ValidAttributeHookNameSniff implements Sniff
+class ValidHookNameSniff implements Sniff
 {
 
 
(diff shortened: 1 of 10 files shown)


coder_sniffer/Drupal/Sniffs/NamingConventions/ValidAttributeHookNameSniff.php
```diff
@@ -0,0 +1,69 @@
+<?php
+/**
+ * \Drupal\Sniffs\NamingConventions\ValidAttributeHookNameSniff.
+ *
+ * @category PHP
+ * @package  PHP_CodeSniffer
+ * @link     http://pear.php.net/package/PHP_CodeSniffer
+ */
+
+namespace Drupal\Sniffs\NamingConventions;
```
klausi: In the naming conventions folder we have PHP names, so I don't think the property string of an attribute belongs there.

Let's make a new folder "Attributes" and just call the class "ValidHookNameSniff"

coder_sniffer/Drupal/Sniffs/NamingConventions/ValidAttributeHookNameSniff.php
```diff
@@ -0,0 +1,69 @@
...
+     * @return void|int Optionally returns a stack pointer. The sniff will not be
+     *                  called again on the current file until the returned stack
+     *                  pointer is reached. Return $phpcsFile->numTokens + 1 to skip
+     *                  the rest of the file.
+     */
+    public function process(File $phpcsFile, $stackPtr)
+    {
+        $tokens = $phpcsFile->getTokens();
+
+        if ($tokens[$stackPtr + 1]['type'] === 'T_STRING'
```
klausi: for all those conditions we should check first with isset() if the next token is there. This will help prevent failures in live coding when the code is just being written.

coder_sniffer/Drupal/Sniffs/NamingConventions/ValidAttributeHookNameSniff.php
```diff
@@ -0,0 +1,69 @@
...
+            && $tokens[$stackPtr + 1]['content'] === 'Hook'
+            && $tokens[$stackPtr + 3]['type'] === 'T_CONSTANT_ENCAPSED_STRING'
+            && str_contains($tokens[$stackPtr + 3]['content'], 'hook_')
```
klausi: we need to make an exception for `hook_info`.

coder_sniffer/Drupal/Sniffs/NamingConventions/ValidAttributeHookNameSniff.php
```diff
@@ -0,0 +1,69 @@
...
+        ) {
+            $hookName = $tokens[$stackPtr + 3]['content'];
+            $phpcsFile->addFixableWarning('Hook name should not start with "hook_" prefix. Hook name used:' . $hookName, $stackPtr + 3,'AttributePrefixHookName');
+            $phpcsFile->fixer->replaceToken($stackPtr + 3, str_replace('hook_', '', $hookName));
```
klausi: before replacing tokens we should check if the fixer is active, by assigning a `$fix` variable on the line before and checking that. See how we do that in other places.

tests/Drupal/bad/BadUnitTest.php
```diff
@@ -384,7 +384,7 @@ protected function getErrorList(string $testFile): array
                 849 => 2,
                 860 => 2,
                 867 => 1,
-                871 => 2,
+                879 => 2,
```
klausi: Since this sniff will likely not interfere with others we should leave BadUnitTest alone and do this in a dedicated test file for this new sniff.

tests/Drupal/bad/bad.php
```diff
@@ -868,4 +868,12 @@ function test30(TestType $a = NULL) {
   echo "Hello";
 }
 
+/**
+ * Implements hook_node_view().
+ */
+#[Hook('hook_node_view')]
+function bad_node_view() {
```
klausi: please also add a test for `hook_info` that is passes in the new sniff test file.

klausi (changes requested):
@RuZniki nice to hear from you :)

Left some comments, please also check the test fails.

coder_sniffer/Drupal/Sniffs/NamingConventions/ValidAttributeHookNameSniff.php
RuZniki: [nicxvan](https://www.drupal.org/u/nicxvan) mentioned that ([link](https://www.drupal.org/project/coder/issues/3499291#comment-15947940)) hook_hook_info cannot be oop, but I decide to add it anyway. Maybe it will be more exceptions in the future.

coder_sniffer/Drupal/Sniffs/NamingConventions/ValidAttributeHookNameSniff.php
RuZniki: I swtiched to `->findNext`, do you think this is enough?

coder_sniffer/Drupal/Sniffs/Attributes/ValidHookNameSniff.php
```diff
@@ -0,0 +1,90 @@
...
+ */
+class ValidHookNameSniff implements Sniff
+{
+
+    /**
+     * List of hooks that should not be fixed.
+     *
+     * @var string[]
+     */
+    public array $hookExceptions = ['hook_info'];
```
RuZniki: Not sure, but typed properties came at php 7.4, in composer `"php": ">=7.2"`. Should I remove type property?

RuZniki:
@klausi I am also happy to get feedback from you, thanks a lot for your review.

I moved sniff to `\Drupal\Sniffs\Attributes\ValidHookNameSniff`.
And moved tests from bad.php to `\Drupal\Test\Attributes\ValidHookNameUnitTest`.
I decide not to test, when name arguments for Hook attribute in wrong order. Example:
```php
#[Hook(method: "userLogin", hook: "hook_user_login")] 
```
I believe this should not be a part of this sniff.

I add special handling for hook_hook_info(). But still not sure about this case.

PR is ready for review.

RuZniki:
I removed special handling for "hook_hook_info()".

coder_sniffer/Drupal/Sniffs/Attributes/ValidHookNameSniff.php
```diff
@@ -0,0 +1,81 @@
...
+        ) {
+            $hookName = $phpcsFile->findNext(T_CONSTANT_ENCAPSED_STRING, ($attributeName + 2));
+            if ($hookName !== false
+            ) {
+                // Remove outer quotes.
+                $hookNameValue = trim($tokens[$hookName]['content'], '"\'');
+
+                if (strpos($hookNameValue, 'hook_') === 0) {
+                    $fix = $phpcsFile->addFixableWarning("Hook name should not start with 'hook_'. Hook name used: $hookNameValue", $hookName, 'HookPrefix');
+                    if ($fix === true && strlen($hookNameValue) > 5) {
```
klausi: the strlen() condition here can be removed, this was already checked before.

We should not add additional conditions with `$fix`, it would be very confusing if the fixer does not always work.

coder_sniffer/Drupal/Sniffs/Attributes/ValidHookNameSniff.php
```diff
@@ -0,0 +1,81 @@
...
+                        // Remove "hook_" prefix.
+                        $hookNameValueFixed = substr($hookNameValue, 5);
```
klausi: we should use this in the error message above like "The hook name should not start with 'hook_', expected '%s' but found '%s".

klausi (changes requested):
Sorry for the delay!

Looks good to me, just 2 minor things inline.

coder_sniffer/Drupal/Sniffs/Attributes/ValidHookNameSniff.php
RuZniki: I tried to avoid case when hook is not finished, and it's just "hook_".
Now I get your point, I will move second clause to above statement. Thanks.

coder_sniffer/Drupal/Sniffs/Attributes/ValidHookNameSniff.php
RuZniki: Thanks, I update error message.
Result example

> The hook name should not start with 'hook_', expected 'node_alter' but found 'hook_node_alter' (Drupal.Attributes.ValidHookName.HookPrefix)

klausi:
Thanks, Drupal core regression tests are failing, but looks unrelated to this issue. Will check in a follow-up!

PR #268, opened 2025-07-13
Exclude DisallowShortOpenTag from .yml files

jonathan1055:
Do not run Generic.PHP.DisallowShortOpenTag on .yml files
Drupal issue https://www.drupal.org/project/coder/issues/3535411

### DIFF
--- a/coder_sniffer/Drupal/ruleset.xml
+++ b/coder_sniffer/Drupal/ruleset.xml
@@ -65,7 +65,7 @@
   <rule ref="Generic.PHP.DeprecatedFunctions"/>
   <rule ref="Generic.PHP.DisallowShortOpenTag">
     <!-- Do not run this sniff on .yml files -->
-    <!-- <exclude-pattern>*.yml</exclude-pattern> -->
+    <exclude-pattern>*.yml</exclude-pattern>
   </rule>
   <rule ref="Generic.PHP.LowerCaseKeyword"/>
   <rule ref="Generic.PHP.UpperCaseConstant"/>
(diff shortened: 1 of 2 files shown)


coder_sniffer/Drupal/ruleset.xml
```diff
@@ -63,7 +63,10 @@
   <rule ref="Generic.NamingConventions.ConstructorName"/>
   <rule ref="Generic.NamingConventions.UpperCaseConstantName"/>
   <rule ref="Generic.PHP.DeprecatedFunctions"/>
-  <rule ref="Generic.PHP.DisallowShortOpenTag"/>
+  <rule ref="Generic.PHP.DisallowShortOpenTag">
+    <!-- Do not run this sniff on .yml files -->
+    <exclude-pattern>**/*.yml</exclude-pattern>
```
klausi: any reason why we need `**`? Other exclusions in this file just use `<exclude-pattern>*.tpl.php</exclude-pattern>`

jonathan1055: Um, no. It's 3 months ago and I can't recall why I did it like that! Maybe I did not look at the rest of the file and just used a pattern that I thought was necessary. Other filepath patterns need a double `**` (eg ESLint ignore, I think). I didn't do too much testing at the time, as I had no idea if you would take on this suggestion. Now that I know it's on, I will add test coverage

jonathan1055: I have removed the double `**` and added a `good/good.yml` file to demonstrate the failure, then the fix.

